PR #19, opened 2026-07-25
Decompose the architecture draft into PR-sized specifications

Krzysztof318:
## What this changes

The architecture draft in `specs/` described the target system, but nothing mapped it onto reviewable units of work. This adds `specs/README.md` as a roadmap index plus 21 specifications covering delivery stages 1 through 4 and the deferred schema and integration-verification work, then updates the draft to reference them.

No production or test code changes. Documentation and specifications only.

## Scope of each specification

One pull request of roughly 1000 changed lines or fewer, including tests and documentation. Each states its roadmap group, the draft stage it serves, its dependencies, an estimated size, and a definition of done. Implementation plans are written per specification when that work starts, not in advance.

| Group | Specifications |
|---|---|
| A — configuration, transport security, resilience | 01 transport security policy · 02 secret reference resolution · 03 resilience pipeline foundation · 04 IMAP session resilience · 05 folder configuration and discovery |
| B — message data enrichment | 06 MIME metadata extraction · 07 metadata persistence and indexes · 08 extracted text and full-text index |
| C — continuous synchronization | 09 per-account supervisor · 10 expunge and flag reconciliation · 11 IMAP IDLE · 12 NOTIFY and CONDSTORE |
| D — read side and MCP | 13 mailbox query read models · 14 content read model · 15 lexical search · 16 MCP hosting and `list_emails` · 17 `get_email_content` · 18 `search_emails` |
| E — schema and infrastructure verification | 19 migration baseline and apply policy · 20 Aspire integration test foundation · 21 IMAP behavior integration tests |

Draft stages 5 through 10 — pgvector, RAG and `ask_mail`, SMTP outbox, OAuth and mTLS, production hardening — are decomposed later, against the code that exists by then.

## Three deliberate departures from the draft's stage order

**Migrations move from stage 1 to the end (specification 19).** Generating migrations while the schema is still growing produces a dozen incremental migrations to review, several partially reverting each other. One baseline migration is reviewed once against the settled schema instead. Specification 07 covers the interval with a Development-only schema bootstrap that specification 19 deletes. Migrations are added through `aspire exec --resource mailmcp-host -- dotnet ef migrations add`.

**Full-text indexing moves from stage 5 to before stage 4 (specification 08).** `search_emails` cannot be built before the index it queries exists.

**Resilience becomes an explicit foundation (specification 03).** The draft asked for bounded jittered backoff and isolated retry state but never named a mechanism, and the repository has resilience only for `HttpClient` via `AddStandardResilienceHandler`. IMAP, SMTP, PostgreSQL, and the future AI providers have no timeout, retry, or circuit breaker at all. Specification 03 establishes named Polly v8 pipelines per dependency class with a single-layer composition rule and an explicit position on the EF Core execution-strategy interaction.

## Two design decisions worth reviewer attention

**Interim MCP security posture (specification 16).** OAuth 2.1 and mTLS are stage 9. Rather than shipping an unauthenticated mailbox endpoint in the meantime, specification 16 keeps the MCP endpoint disabled by default, bound to loopback when enabled, and fails startup on any non-loopback address until the OAuth work removes the gate. Authorization checks are still written at the tool handler so the later work adds identity rather than adding authorization for the first time.

**Integration tests run through Aspire test mode (specifications 20 and 21).** `DistributedApplicationTestingBuilder` drives the existing `AppHost` app model rather than a parallel container topology, so the orchestration under test matches the one developers run. Specification 20 pays off the PostgreSQL mapping and constraint verification that ADR 001 defers; specification 21 proves the `\Seen` invariant against a real IMAP server, which no substitute-based unit test can do. Specification 20 also updates `AGENTS.md`, which currently forbids integration-test projects outright.

## Draft updates

- Implementation status and delivery stages now reference the specifications.
- Technology baseline gains rows for Polly v8 and Aspire test mode.
- New section 17.1 describes the resilience pipeline model, the single-layer rule, and the terminal-versus-transient classification.
- Section 6.2 replaces vague container language with the Aspire app-model approach.
- `docs/README.md` points at `specs/` and states the distinction: a specification is intent, a docs page is fact.

## Verification

Markdown only, no build or test impact. All relative links in `specs/README.md` resolve to existing files. The Aspire APIs named in specifications 19 through 21 (`aspire exec`, `DistributedApplicationTestingBuilder`, `Aspire.Hosting.Testing`) and the Polly registration model in specification 03 (`AddResiliencePipeline`, `ResiliencePipelineProvider`) were verified against current official documentation before being written down.


Krzysztof318:
## Update after owner review

**Stage 4 may expose the MCP endpoint externally.** The loopback gate in specification 16 is removed — this segment is development work and release is far off. What remains is explicitness rather than enforcement: the endpoint is disabled by default, startup warns unambiguously when it is enabled without transport authentication, and `docs/operations/` records that the pre-stage-9 posture is unauthenticated. Authorization at the tool handler is unchanged, so stage 9 adds transport identity to an existing decision.

**The HTML sanitizer evaluation for specification 14 is now complete** rather than deferred to implementation time.

**Selected: `HtmlSanitizer` 9.0.967** — MIT, published 2026-07-21, 120.7M downloads, AngleSharp-based parsing, allow-list model for tags, attributes, CSS properties, and URI schemes. Targets `net8.0`, which is compatible with .NET 10.

Three consequences worth knowing before the work starts:

1. **It pins AngleSharp to exactly `0.17.1`** plus `AngleSharp.Css` `0.17.0`. AngleSharp itself is at 1.5.2, so this forecloses referencing AngleSharp 1.x anywhere in the solution — the exact pin cannot be satisfied alongside it. Nothing else in MailMcp needs AngleSharp, so it costs nothing today, but it is recorded so a later change does not find out via build failure.
2. **`LICENSES.md` needs three entries**, not one: `HtmlSanitizer`, `AngleSharp`, `AngleSharp.Css`. All MIT.
3. **Security history:** CVE-2026-25543 / GHSA-j92c-7v7g-gj3f (moderate, 2026-02-03) was a bypass where `<template>` contents went unsanitized, exploitable only when `template` and `shadowrootmode` were explicitly allowed. Fixed in 9.0.892, so 9.0.967 carries the fix, and the default configuration was never affected. The directional lesson is in the spec: keep the allow-list minimal, never add `template`, and watch the pinned version for advisories.

Rejected, with reasons recorded: **AngleSharp 1.5.2 alone** (parser, not sanitizer — using it means hand-writing allow-list, CSS filtering, and URI-scheme handling), **HtmlAgilityPack 1.12.4** (same, and last released 2025-10-03), **`Microsoft.Security.Application` / AntiXSS** (legacy, .NET Framework only, unmaintained).

**Specification 08 gains a related decision.** HTML-to-text derivation uses `MimeKit.Text.HtmlTokenizer`, already available through the pinned MailKit 4.17.0, so it adds no dependency. Given the AngleSharp pin above, this keeps a second HTML stack out of the solution rather than trading one version conflict for another.

The specification index gains a dependency summary table covering all five specifications that add one.

Sources: [HtmlSanitizer on NuGet](https://www.nuget.org/packages/HtmlSanitizer), [mganss/HtmlSanitizer](https://github.com/mganss/HtmlSanitizer), [GHSA-j92c-7v7g-gj3f](https://github.com/mganss/HtmlSanitizer/security/advisories/GHSA-j92c-7v7g-gj3f), [AngleSharp on NuGet](https://www.nuget.org/packages/AngleSharp), [AngleSharp.Css on NuGet](https://www.nuget.org/packages/AngleSharp.Css), [HtmlAgilityPack on NuGet](https://www.nuget.org/packages/HtmlAgilityPack), [MimeKit HtmlTokenizer](https://mimekit.net/docs/html/T_MimeKit_Text_HtmlTokenizer.htm)

Krzysztof318:
## Response to Codex review

All five findings verified against `AGENTS.md` and ADR 0002. All five hold, all five are fixed in da032e2.

### P1 — Keep automatic migrations out of production startup (spec 19)

**Valid.** `AGENTS.md:212` is unconditional: "Do not run destructive or automatic production migrations during ordinary host startup." An opt-in that an operator *can* enable in Production still permits exactly that, so the setting was the problem, not its default.

Replaced with a hard environment boundary rather than a setting: Development may apply pending migrations at startup; every other environment verifies the schema matches the expected migration set and fails startup when migrations are pending. Applying is a deployment step via `aspire exec … -- dotnet ef database update`.

Draft section 17 stated the weaker policy and is updated too, with the reasoning made explicit: a starting instance that mutates schema can race a second instance, can apply an unreviewed destructive change, and leaves the operator no point at which to take a backup. The draft had labelled its own policy an "arbitrary initial policy", so the repository rule wins.

### P1 — Give each remapped remote folder a distinct occurrence scope (spec 05)

**Valid, and the most useful finding here.** UIDVALIDITY is unique within a mailbox, not across a server, so two unrelated folders can carry the same value. Keying persistence on account plus alias meant a repointed alias could inherit the old folder's checkpoint, and its `LastSeenUid` would skip every message below that UID in the new folder — silently, permanently, with no error anywhere.

The folder row now carries a resolution generation and the occurrence scope is keyed on it. Repointing an alias starts a new generation with no checkpoint, so the new folder synchronizes from the beginning under its own identity; occurrences from the previous generation stay attributable to the folder they came from. The audit event now names both remote paths and the new generation.

Added as a named test in the spec: repointing to a folder whose UIDVALIDITY equals the old one must not skip messages. That is the case a generation-free design gets wrong, and it needs a test that would catch its removal.

### P2 — Keep secret resolution at the host boundary (spec 02)

**Valid.** ADR 0002's scope exclusions permit consuming already-bound secret values at the host boundary and forbid normalizing broad secret access into application code. An `ISecretResolver` visible to `Application` gives every use case the ability to ask for any secret by name, which is that.

Contract and per-scheme adapters moved to `Infrastructure`, invoked once by `Host` during startup validation. Application and domain code receive only the resolved, narrowly scoped settings each operation needs. Added an architecture test asserting no secret-resolution type is reachable from `Application` or `Domain`.

### P2 — Assign trusted-CA loading to an implementation specification (spec 01)

**Valid — the work was genuinely orphaned.** Spec 01 promised spec 02 would wire it, spec 01's own out-of-scope excluded it, and spec 02 said nothing about it, so `AGENTS.md:165` ("Support private servers through explicit trusted CA configuration") would never have been satisfied.

Assigned to specification 02, which owns deployment-provisioned material and builds the reference-resolution mechanism a trust anchor arrives through. Spec 01 now defines only the configuration shape and its validation; spec 02 gains the loading, the installation into MailKit's validation path, tests for a valid anchor, a malformed anchor failing startup, and the absence of any path that disables validation.

### P2 — Verify SQL parameterization in the infrastructure adapter (spec 15)

**Valid.** An in-memory repository generates no SQL, so a metacharacter test there would pass against a vulnerable adapter and prove nothing.

Moved to `Infrastructure.UnitTests`, which observe the generated command and verify query text arrives as a parameter rather than concatenated SQL, including full-text operator cases. Specification 20 gains the end-to-end confirmation against real PostgreSQL alongside the existing GIN index-usage check. The spec now states plainly why the assertion cannot live at the application level, so it does not drift back.

Issue #44, opened 2026-07-25
Spec 10 — Remote expunge and flag reconciliation


Krzysztof318:
**Roadmap group:** C — continuous synchronization · **Draft stage:** 3
**Specification:** [`specs/10-remote-expunge-and-flag-reconciliation.md`](https://github.com/Krzysztof318/MailMcp/blob/main/specs/10-remote-expunge-and-flag-reconciliation.md)
**Depends on:** #41, #43 · **Size:** ~700 lines including tests and documentation

### User stories
**As a mailbox owner** I want a message I delete on the server to disappear from MailMcp within one reconciliation cycle, **so that** the local copy does not resurrect mail I removed.

**As a mailbox owner** I want deleted content and everything derived from it — extracted text, index entries, later embeddings — removed together after the retention grace period, **so that** deletion actually means deletion.

**As an operator** I want a UIDVALIDITY change to trigger controlled reconciliation rather than mass local deletion, **so that** a server-side renumbering does not wipe my local mailbox.

### Acceptance
- A message deleted on the server is marked deleted locally within one reconciliation cycle of its window.
- Content and all derived data are removed after the grace period; the tombstone survives per its own retention setting.
- A UIDVALIDITY change never causes mass local deletion.
- `docs/features/imap-synchronization.md` documents reconciliation windows, retention settings, and the audit events.

PR #132, opened 2026-07-29
Prove the Seen-preserving invariant against a real IMAP server (#55)

### LINKED ISSUE #55: Spec 21 — IMAP behavior integration tests
**Roadmap group:** E — schema consolidation and infrastructure verification · **Draft stage:** deferred integration phase
**Specification:** [`specs/21-imap-behavior-integration-tests.md`](https://github.com/Krzysztof318/MailMcp/blob/main/specs/21-imap-behavior-integration-tests.md)
**Depends on:** #54 · **Size:** ~600 lines including tests and documentation

### User stories
**As a mailbox owner** I want the "synchronization never marks mail as read" claim proven against a real IMAP server by reading the flag back, **so that** the repository's most-repeated promise stops resting on a substituted port.

**As a reviewer** I want a `\Seen` regression to fail with a message naming the message and the observed flag state, **so that** the invariant most likely to break silently during refactoring breaks loudly instead.

**As the owner** I want the server image selected, license-reviewed, and its evaluation written down, **so that** replacing it later is an informed decision.

### Acceptance
- The `\Seen` invariant proven against a real server: full synchronization of unread messages and content fetch both leave every remote flag unchanged, read back after the run.
- Repeated synchronization of the same `(account, folder, UIDVALIDITY, UID)` is idempotent; a UIDVALIDITY change triggers controlled reconciliation without mass local deletion; an expunged message is detected by the #44 reconciliation pass.
- The transport security policy from #35 rejects a clear-text mechanism on an unencrypted channel against a server that offers one.
- Where supported: IDLE delivers a notification triggering exactly one synchronization pass, and CONDSTORE reconciliation reaches the same end state as the full-window path.
- The chosen image, its tag, and its license recorded in `LICENSES.md` with the selection rationale documented.

> Selection criteria in order: UIDVALIDITY semantics and PEEK fetch at minimum, ideally IDLE / NOTIFY / CONDSTORE; scriptable enough to seed messages and read back flag state; license and image terms verified from official sources. If no candidate can observe flag state, that finding is documented and the scope reduces to what stays verifiable.

Krzysztof318:
Closes #55

Specification 21 exists because the repository makes one claim constantly and has never proven it: that synchronization does not mark mail as read. Every unit test substitutes the mailbox port, so all of them prove that application code *asks* for the seen-preserving fetch and none of them can say what the resulting IMAP commands did to a server.

## What the suite now proves

The integration-testing topology starts a real IMAP and SMTP server beside PostgreSQL. The suite seeds mail over SMTP and reads flag state back over its own IMAP connections, so an observation reports the server's answer rather than the adapter's belief about it.

Six tests, deliberately few — an integration test costs container start-up and real I/O on every run:

| Test | What only a real server can establish |
|---|---|
| `MarkSeenAsync_OnADeliveredMessage_IsObservedByTheSuitesOwnConnection` | The control. An assertion that a flag is unset proves nothing unless the same observation would report it set. |
| `FetchEmailContentWithoutSettingSeenAsync_AgainstARealServer_LeavesEveryRemoteSeenFlagUnset` | The invariant, over both operations a session performs against a folder. |
| `SynchronizeAsync_OverAFreshlySeededFolder_...` | A whole run against a real server and a real database leaves every flag untouched. |
| `SynchronizeAsync_OverAFolderAlreadySynchronized_StoresNothingTwiceAndKeepsTheCheckpoint` | Idempotency, read from what the first run persisted. |
| `SynchronizeAsync_AfterTheFolderWasRecreatedWithANewUidValidity_...` | A genuine UIDVALIDITY change restarts the folder without deleting stored mail. |
| `ReadPendingMigrationIdentifiersAsync_...` | Unchanged; joined the shared collection. |

A `\Seen` regression fails with a message naming the message and the state observed, not a generic assertion failure, because this is the invariant most likely to break silently during a later refactoring.

## Server selection

Both candidates were driven by hand before either was wired in.

**smtp4dev** (BSD-3-Clause) preserves `\Seen` under `BODY.PEEK` and sets it under a non-peek fetch, so the invariant itself is observable against it. Rejected on two counts: its INBOX reports a hard-coded UIDVALIDITY that nothing can change, making that scenario permanently unverifiable, and `UID SEARCH UID 1:*` exhausts the container's memory and kills the process — a shape MailMcp never sends, but worth recording.

**GreenMail** (Apache-2.0) was selected. Each folder carries a real UIDVALIDITY, so replacing one produces a genuine change rather than a simulated one.

## The production change this exposed

Neither server advertises a SASL mechanism from MailMcp's allow-list — smtp4dev advertises none at all, GreenMail only `AUTH=XOAUTH2`. Both authenticate with the IMAP `LOGIN` command, which RFC 3501 leaves as the client's last resort and which `MailKitTransportSecurityMapping` refused outright. That refusal made every RFC-conformant server without an `AUTH=` capability unreachable by MailMcp, with no remedy available to an operator.

The adapter now permits the fallback exactly when the account's policy already permits a clear-text mechanism — the same exposure, already opted into, and on an unencrypted channel still gated by the separate opt-in the policy requires before it can be built. An allow-list of challenge-response mechanisms alone still ends the attempt with `MailAuthenticationMechanismUnavailableException`. Four unit tests cover the new branch and the one that must keep failing.

## Test rules added to `tests/AGENTS.md`

- Decide how many integration tests a class earns before writing any, and write only what real infrastructure can prove.
- Tests sharing a resource go in one collection with parallelization off; `OrchestratedInfrastructureCollectionDefinition` is that collection.
- Ordering is allowed in this suite and nowhere else, for a seeding step followed by a reading step.
- An assertion of absence carries a control that produces the thing and observes it.

## Out of scope, and why

Expunge detection waits on #44 and CONDSTORE on the specification-12 work; GreenMail advertises no CONDSTORE. IDLE is available but unexercised, because specification 11 is not implemented. This is recorded in `specs/21`, which anticipated the reduction.

## Verification

`scripts/verify-full.sh` exits 0: 842 unit tests, 96.49% aggregate coverage, formatting clean. `scripts/run-integration-tests.sh` runs 6 of 6 green; the informational integration-coverage report over the 22 marked classes rises from 26.9% to 71.7%.


src/Infrastructure/Infrastructure.csproj
```diff
@@ -35,5 +35,9 @@
 
   <ItemGroup>
     <InternalsVisibleTo Include="MailMcp.Infrastructure.UnitTests" />
+    <!-- The integration suite reads back what a synchronization run persisted, and the EF Core model it has to read is
```
Krzysztof318: Komentarz jest nadmiarowy, usuń go

tests/IntegrationTests/IntegrationTests.csproj
```diff
@@ -5,6 +5,11 @@
          host so that infrastructure's hosted lifecycle events run, which is what makes the PostgreSQL connection string
          available to anything that opens a connection. -->
     <PackageReference Include="Microsoft.Extensions.Hosting" />
+    <!-- Referenced directly rather than taken transitively through Infrastructure, because the suite speaks IMAP and
```
Krzysztof318: Komentarz nadmiarowe, usun go

tests/IntegrationTests/SyntheticMailAccount.cs
```diff
@@ -0,0 +1,55 @@
...
+namespace MailMcp.IntegrationTests;
+
+/// <summary>The one throwaway mailbox the orchestrated mail server serves, as the adapter's two ports see it.</summary>
+/// <remarks>
+/// <para>
+/// The suite supplies these ports itself rather than composing the host's configuration-bound ones, for the same reason
+/// it does not start the host resource: what is under test is the mail adapter against a real server, not how a
+/// composition root binds an options section. Both implementations are small enough that stating them here is clearer
+/// than reaching into another assembly's internals for them.
+/// </para>
```
Krzysztof318: Uporządkuj klasy w integrationtest a nie wszystko w jednym katalogu

src/Infrastructure/Infrastructure.csproj
Krzysztof318: Usunięte w 7231d78 — zostało samo `<InternalsVisibleTo Include="MailMcp.IntegrationTests" />`, tak jak sąsiedni wpis dla testów jednostkowych.

tests/IntegrationTests/IntegrationTests.csproj
Krzysztof318: Usunięte w 7231d78.

tests/IntegrationTests/SyntheticMailAccount.cs
Krzysztof318: Uporządkowane w 7231d78. Katalogi nazwane tym, czemu służą, z przestrzeniami nazw dopasowanymi do nich:

```
Orchestration/   MailMcpOrchestrationFixture, OrchestratedMailMcpServices,
                 OrchestratedMailServerEndpoints, SyntheticMailAccount,
                 OrchestratedInfrastructureCollectionDefinition
Mailbox/         OrchestratedMailbox, ObservedEmail, RemoteSeenFlagAssertion,
                 MailboxStateSequenceOrderer, MailboxStateStepAttribute
Persistence/     OrchestratedDatabaseSchemaTests
Synchronization/ RemoteSeenFlagPreservationTests, OrchestratedMailboxSynchronizationTests
AssemblyFixtures.cs
```

Podział przebiega tam, gdzie leży odpowiedzialność: `Orchestration` to wszystko, co uruchamia zorkiestrowaną infrastrukturę i do niej sięga, `Mailbox` to niezależny świadek po stronie IMAP wraz z jego asercją i kolejnością, a dwa pozostałe katalogi to same testy pogrupowane po tym, co weryfikują. Przy okazji `MailboxStateStepAttribute` i `ObservedEmail` dostały własne pliki, bo dzieliły plik z innym typem.

Issue #143, opened 2026-07-30
Authenticate MCP users through provider-neutral OAuth 2.1 and OpenID Connect


Krzysztof318:
## Context

**No numbered specification under `specs/` backs this issue.** This is later, provider-neutral authentication for the inbound MCP Streamable HTTP endpoint. #140 deliberately limits the first authentication stage to API keys, explicit unauthenticated mode, CORS, and mTLS, and leaves OAuth 2.1 for later work. #78 is unrelated outbound OAuth for IMAP and SMTP providers.

MailFathom must act as an OAuth 2.1 protected resource and delegate user authentication and token issuance to an external authorization server. Keycloak is the required self-hosted compatibility case, but the contract must be standards-based so other compliant OAuth 2.1 or OpenID Connect providers can be configured without provider-specific authentication code or identity types leaking into `Domain` or `Application`. MailFathom does not become an authorization server and does not store user passwords, authorization codes, or refresh tokens.

Current Keycloak documentation reports partial compliance with the latest MCP authorization profile because Resource Indicators (RFC 8707) are not yet handled directly; its documented audience-mapper setup can still bind tokens to the MCP resource. Compatibility must never be achieved by disabling audience validation.

Official references:

- [MCP authorization specification](https://modelcontextprotocol.io/specification/2025-11-25/basic/authorization)
- [OpenAI authentication for MCP apps](https://developers.openai.com/apps-sdk/build/auth)
- [Keycloak MCP authorization-server integration](https://www.keycloak.org/securing-apps/mcp-authz-server)

## User stories

**As an operator with an existing identity platform** I want to connect MailFathom to Keycloak or another standards-compliant provider, **so that** MCP users authenticate through infrastructure I already operate instead of receiving MailFathom-specific credentials.

**As an MCP user** I want the client to open the provider's normal interactive sign-in and consent flow, **so that** personal, organizational, MFA, and conditional-access policies remain owned by that provider.

**As a security owner** I want issuer, audience, scope, lifetime, and signature checks enforced identically for every provider, **so that** adding provider compatibility cannot weaken the MCP resource boundary.

### Provider-neutral configuration

- Add an explicit `OAuth` MCP authentication mode alongside the modes from #140. Missing or invalid configuration never falls back to `None`, API keys, or a less strict provider profile.
[...]

Issue #139, opened 2026-07-30
Adopt MailFathom as the project identity before the first public release


Krzysztof318:
## Context

**No numbered specification under `specs/` backs this issue.** The owner has selected `MailFathom` as the product and project name. The rename must land before the first public release so the public repository, runtime, configuration, documentation, and contributor experience expose one identity from the outset.

The repository currently carries `MailMcp` across solution and file names, namespaces and assemblies, branded type names, configuration and environment-variable prefixes, Aspire and persistence resource names, scripts, tests, documentation, runtime service identity, and repository metadata. MailFathom is unreleased, so this is a clean contract move: every caller and durable document changes together, with no compatibility aliases, fallback keys, migration path, or product-facing rebranding narrative.

The desired end state is a current tree and active repository that read as MailFathom, not as a product documenting a transition from an earlier name. Git history and already-closed GitHub discussions remain historical records and are not rewritten.

## User stories

**As a first public user** I want the repository, executable behavior, configuration examples, and MCP server identity to use `MailFathom` consistently, **so that** I encounter one product name and can search for it reliably.

**As a contributor** I want solution names, assemblies, namespaces, tests, scripts, and operational resources to share the same root identity, **so that** a clone builds and runs without stale names or misleading mixed terminology.

**As the project owner** I want the first public release to present MailFathom as the project identity without “formerly known as” or rebranding prose, **so that** temporary development history does not become part of the product story.

## Acceptance

- The active GitHub repository name and description, roadmap project title, default clone URLs, and other editable repository metadata use `MailFathom`/`mail-fathom` consistently.
- Every tracked file and directory name that contains the old brand is renamed, including the solution file and any branded source or test type whose name represents the product rather than the MCP protocol itself.
- Assembly names, root namespaces, C# namespaces and type names, `InternalsVisibleTo` entries, coverage filters, solution references, and build or test scripts use `MailFathom` consistently.
- Runtime-facing identity is renamed wherever it is product identity: MCP server implementation information, readiness responses, application and telemetry service names, structured-log categories, user-agent or client identity strings, executable/package/container names, and deployment artifacts.
- Configuration sections, environment-variable prefixes, secret names, Aspire resources, database/resource identifiers, test fixtures, example domains and addresses, and persisted schema identifiers carrying the old brand are changed in full. Because the project is pre-release, the baseline is regenerated or renamed directly where required; no compatibility fallback or data migration for an unreleased contract is added.
[...]

Issue #123, opened 2026-07-29
Build the canonical secure multi-architecture container image


Krzysztof318:
## Context

**No numbered specification under `specs/` backs this issue.** It is a release obligation: MailFathom treats the container as a first-class deployment shape, and there is no canonical application image yet.

#158 owns the checked-in Dockerfile and the Compose/Helm deployment assets. This issue owns the image contract that Dockerfile must satisfy: reproducibility, runtime contents, architecture support, identity, least privilege, health behavior, and licensing. #156 owns publication and deliberately separates its registries: stable releases go only to Docker Hub, while manually triggered nightly builds go only to GHCR.

The base image is a licensing item in its own right. `AGENTS.md` treats container images as separately reviewable from SDK licenses, so every `mcr.microsoft.com/dotnet/*` base and anything else layered into the image gets an entry in the third-party license register with any required attribution or NOTICE handling.

## User stories

**As an operator** I want to run an official MailFathom image without installing a .NET SDK, **so that** evaluation and deployment use the same reviewed artifact.

**As a security-conscious operator** I want the image to run as a non-root user with no secrets baked in, **so that** deploying a mailbox-reading service does not grant it unnecessary privilege.

**As an operator on ARM hardware** I want an image that runs on my machine, **so that** self-hosting on small always-on hardware is supported.

## Acceptance

- The checked-in Dockerfile from #158 builds the canonical image reproducibly from a clean clone with no developer-machine-only step and no second, drifting build definition.
- Build and runtime stages are separated. The runtime image contains only the published application and explicitly required runtime assets; it contains no SDK, source tree, repository history, test artifacts, credentials, certificates, connection strings, or build caches.
- Every base image is pinned by digest or by an explicitly managed immutable version, verified from the official upstream source, and recorded in the third-party license register with its exact reference and required attribution or NOTICE handling.
- The image runs as a non-root user, supports a read-only root filesystem with explicit writable locations, drops unnecessary privileges, and contains no embedded secrets. Configuration comes from environment variables and mounted secret files consistent with the secret-provisioning contract.
- Multi-architecture manifests support `linux/amd64` and `linux/arm64`, and both platform variants are smoke-tested rather than only declared.
- The built application carries the version/revision identity from #119. Stable Docker Hub tags follow #116; GHCR nightly tags and channel metadata follow #156. The same image contract applies in both channels without implying that a nightly is a release.
[...]

Issue #113, opened 2026-07-29
Publish the project's own license and carry it into the build metadata and the artifacts


Krzysztof318:
## Context

**No numbered specification under `specs/` backs this issue.** It is a first-release obligation: MailFathom has no license of its own.

What exists today is `LICENSES.md`, which #149 will rename to `THIRD_PARTY_LICENSES.md`. That file records what MailFathom *consumes* and under which terms; it is not a grant of rights to anyone who receives MailFathom. There is no project `LICENSE` or `NOTICE` at the repository root, the shared build metadata does not yet declare the selected project license, and distributed artifacts do not yet carry a complete project-license and attribution bundle.

## Owner decision

The owner has selected the **Apache License, Version 2.0**, with SPDX identifier `Apache-2.0`, for the MailFathom source code and first public release.

The owner also requires a root **`NOTICE`** that identifies **Krzysztof Kasprowicz as the original author and creator of MailFathom**. Apache-2.0 section 4(d) requires derivative distributions of a work that includes a NOTICE file to preserve a readable copy of the relevant attribution notices. The NOTICE is informational, does not modify the Apache-2.0 terms, and must not imply ownership of later contributions written by other copyright holders.

GitHub must recognize the project license independently of build metadata or README prose. Its Licensee-based detection compares the repository's root `LICENSE` file with known license texts, so `LICENSE` remains the unmodified Apache-2.0 text while copyright and original-author attribution stay in `NOTICE`, README, source headers, and build metadata.

Apache-2.0 is a permissive open-source license that preserves commercial use and distribution while adding an explicit patent license and notice obligations. The repository and every produced artifact must apply that decision consistently. This issue no longer compares or selects between license candidates.

Source files already use an EditorConfig-enforced copyright header. The required header must now contain a second, separate line identifying Apache-2.0. Microsoft documents `file_header_template` as the EditorConfig option for IDE0073, with `\n` representing line breaks; `dotnet format` applies and verifies that configured style.

## User stories

**As the project owner and original author** I want the Apache-2.0 decision and my original-authorship attribution preserved in source and derivative distributions, **so that** forks can evolve independently without erasing where MailFathom originated.

**As a first external user** I want GitHub and the repository files to show clearly what I am permitted to do with MailFathom and who originally created it, **so that** I can evaluate and redistribute it without guessing.

[...]

PR #173, opened 2026-07-31
Rename and revalidate the third-party license register

### LINKED ISSUE #149: Revalidate and rename the third-party license register
## Context

**No numbered specification under `specs/` backs this issue.** This is a first-release publication obligation. The root `LICENSES.md` is not the MailFathom project license and is not a distribution-ready NOTICE bundle; it is a maintained register of third-party packages, tools, services, protocols, container images, templates, and externally sourced material reviewed for compatibility.

The generic filename becomes ambiguous once #113 adds the project's own `LICENSE`. Rename the register to `THIRD_PARTY_LICENSES.md`, which describes its actual scope more clearly than `EXTERNAL_LICENSES.md` and does not imply that it already contains every full license or notice that must accompany a distributed artifact.

The existing register has grown throughout development and must be revalidated against what the repository actually uses before it becomes public. This issue owns that complete review and the rename; #113 remains responsible for choosing the project's own license and for the packaging mechanism that emits required notices.

## User stories

**As an external user** I want the project license and third-party license register named unambiguously, **so that** I can distinguish rights granted for MailFathom from obligations inherited from its dependencies.

**As a release owner** I want every third-party entry revalidated against the versions and components actually used, **so that** the first public release does not rely on stale, missing, or planned-only licensing assumptions.

**As a maintainer** I want every repository reference updated to the canonical filename, **so that** documentation and completion gates continue to enforce the register rather than silently pointing to a removed path.

## Acceptance

- Rename root `LICENSES.md` to `THIRD_PARTY_LICENSES.md` in one change and preserve its Git history as a rename rather than creating an unrelated replacement file.
- Update the title and introduction to distinguish the third-party review register from the project's own `LICENSE` and from any generated `THIRD-PARTY-NOTICES` artifact shipped with binaries or containers.
- Inventory the repository's actual external surface, including centrally pinned direct and transitive packages, development and test tooling, container images, GitHub Actions, vendored or template-derived source, generated assets, hosted services, provider APIs, protocols, models, and any externally copied examples.
- Reconcile that inventory with the register: add missing entries, update stale names and exact versions, remove or clearly separate components that are only future ideas, and identify entries whose reviewed source no longer matches the component in use.
- Revalidate every retained entry using current authoritative upstream sources. Record the exact component or version, SPDX license expression or precise custom terms, upstream source, commercial and open-source compatibility decision, attribution or NOTICE requirements, redistribution conditions, and any separately relevant service, telemetry, model, trademark, or data-processing terms.
- Distinguish runtime or redistributed components from build-only, development-only, test-only, and externally hosted dependencies so release obligations are not inferred from planned use alone.
- Resolve every incompatible, unknown, unverified, or conditionally licensed finding before release, or create a blocking narrowly scoped issue that names the owner decision required. Do not silently retain an `Allowed` decision when its evidence cannot be reproduced.
- Update every exact and prose reference to `LICENSES.md` across source, tests, `README`, `docs/`, `specs/`, `AGENTS.md`, companion instruction files, repository-owned skills, scripts, workflows, templates, and issue-facing command output. Verify that the old filename has no remaining live references except deliberate historical text.
- Update the `$check-docs-licenses` completion gate and any other automated validation to read and report the new canonical path.
- Cross-link `THIRD_PARTY_LICENSES.md` with the project `LICENSE` introduced by #113 without duplicating either document's role.
- Record a privacy-safe review summary and the inventory/revalidation commands used so the review can be repeated before later releases.

## Dependencies

- #113 owns the MailFathom project-license decision, root `LICENSE`, and packaging of required notices.
- #120 determines which process artifacts remain in the public tree and therefore remain in the final reference sweep.
- #123 must consume the verified redistribution and notice obligations when building the release container image.
- This issue must complete before the repository becomes public under #121.

## Out of scope

Choosing MailFathom's own license, adding or upgrading dependencies solely to simplify the review, generating the final distribution notice bundle owned by #113, or deciding whether a future component should be implemented.

Krzysztof318:
Closes #149

`LICENSES.md` is not MailMcp's license and never was. It is the register of third-party components reviewed for compatibility with both distribution models, and the name stops saying so the moment #113 adds a `LICENSE`. This renames it to `THIRD_PARTY_LICENSES.md` and revalidates every entry against what the repository actually uses.

## The rename

The first commit is the move alone, with the register's content untouched, so the rename stays visible as a rename. Every live reference follows it: `AGENTS.md`, `NuGet.config`, the `$check-docs-licenses` gate, three `docs/` pages, ADR 0002's path reference (owner-approved, path only), and seven files under `specs/`. The six dated files under `docs/superpowers/` keep the old name deliberately — they record what a past change did, and rewriting them would make that record wrong.

Note for merging: the second commit rewrites enough of the file that similarity-based rename detection no longer fires on the combined diff. A squash merge will therefore show a delete and an add. The two-commit split is what preserves `git log --follow` if the branch is merged or rebased instead.

## What the revalidation found

The register had grown one row at a time alongside the code, so it recorded the direct pins and little else. Exposure was a phrase inside a row rather than the structure, which is what let a release obligation and a build-time dependency sit side by side looking alike.

- **Google.Protobuf 3.30.2 is BSD-3-Clause and is redistributed.** It reaches the artifact through the OTLP exporter. The register recorded OpenTelemetry as Apache-2.0 only, so this notice obligation appeared nowhere. It is the finding the register exists to surface, and it was the one it missed.
- **34 of the 60 redistributed packages were not named.** The identity-model stack, `Microsoft.ML.Tokenizers`, `Microsoft.Extensions.VectorData.Abstractions`, BouncyCastle, the Npgsql satellites, the OpenTelemetry SDK internals, `Pgvector`, `ModelContextProtocol.Core`, and the rest of what the host's graph pulls in.
- **`Microsoft.Extensions.AI.Abstractions` ships as 10.8.3, not the pinned 10.8.1.** `ModelContextProtocol.Core` 2.0.0 requires 10.8.3 and a central pin is a floor. Both are MIT, so nothing is blocked, but the register now records the version an artifact carries rather than the version a file requests. No pin was changed; that would be a dependency decision, not a review.
- **Castle.Core 5.1.1 and the Microsoft Testing Platform packages were missing** from the test inventory, along with the Application Insights client the platform's telemetry extension carries. The telemetry is recorded as a data-processing note with its opt-out variables, not as a licensing problem.
- **Auth0 was recorded as the default identity provider.** It is not one: #143 specifies a provider-neutral design and no Auth0 package is referenced. It is gone from the file entirely.
- **A duplicate, version-less OpenTelemetry row was removed.** It described the same components as the pinned row without naming versions.

No incompatible, unknown, or unverified license was found. Everything resolves to MIT, Apache-2.0, BSD-3-Clause, or the PostgreSQL License. Nothing blocks the first release on licensing grounds.

## How the boundaries were decided

Not by reading the project files and guessing:

- **What ships** is the publish output. `dotnet publish src/Host -c Release` emits 66 assemblies, 60 of them third-party. The redistributed section lists exactly those 60 — verified mechanically in both directions.
- **What only builds** is derived by reachability over the host's dependency graph. Every package in the EF Core Design toolchain is reachable only through that reference, and a publish emits none of them.
- **What the shared framework supplies** is what the host's own graph prunes: 22 packages that appear in the project locks and in no publish output.
- **License terms** come from each package's own published `.nuspec` on nuget.org — 141 packages across the closure and the central pins, zero fetch failures — and from the GitHub license API or upstream `LICENSE` text for everything that is not a NuGet package. pgvector's `NOASSERTION` was resolved by reading the file: it is the PostgreSQL License verbatim with added copyright lines.

The commands are recorded in the register itself, so the next release repeats the review instead of reconstructing it.

Also newly recorded because they are genuinely used and were absent: the Aspire AppHost SDK and its runtime-identifier packages, `Aspire.ProjectTemplates` as the origin of the vendored service-defaults file, Docker Engine and the Docker Desktop subscription threshold, and the three hosted services the repository calls.

## From review

The register described software nobody uses — a section of evaluated, deferred, and rejected candidates, plus a narrative of what the previous revision got wrong. Both are gone, and the rule is now in `AGENTS.md` and the licensing gate rather than only in the file it governs: a row asserts a completed review of software that exists here today, so a planned or rejected candidate keeps its reasoning in the specification, ADR, or issue that owns it, and where the review stood before is what version control is for. The forward-looking half survives as `Open obligations`.

Three findings from Codex were correct and are fixed: the hosted-service rows now cite official terms URLs and the conditions that bind rather than naming documents generically; the Aspire service-defaults row no longer reads a missing copyright header in generated output as a waiver of MIT's notice condition; and `Microsoft.Extensions.Compliance.Abstractions` 10.5.0, which `src/AI` resolves alongside the host's 10.6.0, is recorded. That last one was swept across the whole closure rather than fixed in place — eleven packages resolve at two versions, and the other nine were already covered.

One thread stays open: pinning the CI actions to commit SHAs. It is a sound supply-chain point about `.github/workflows/`, which this pull request does not touch, and the register already records each `v<major>` as a moving tag rather than claiming otherwise.

## Verification

`scripts/verify-full.sh` — green. 1481 unit tests, 0 failed, aggregate line coverage 96.99% against the 85% gate, formatting clean over 574 files, workflow contract suite passed. No production code changed, so no unit test was added; the register's completeness was checked mechanically instead — every centrally pinned package and every lock-file entry is named, 0 unnamed of 141.

`$check-docs-licenses`: pass / pass.

## Out of scope

Choosing MailMcp's own license and generating the notice bundle (#113), the release container image that must carry those notices (#123), and moving any pin. The `LICENSE` cross-link is written as a placeholder pointing at #113, because the file does not exist yet.

The issue text calls the product MailFathom. That rename is #139 and has not landed, so this change keeps the MailMcp naming the repository still uses everywhere.


THIRD_PARTY_LICENSES.md
```diff
@@ -0,0 +1,200 @@
...
+These are called, not linked. None is a redistributable dependency, and none supplies code that ships.
+
+| Service | Use in this repository | Terms that apply | Decision |
+|---|---|---|---|
+| nuget.org | The only package source `NuGet.config` declares, after clearing the machine's inherited source list | NuGet.org terms of use, plus the per-package licenses recorded above | Allowed. Adding a second source is a licensing and supply-chain decision and must be recorded here. |
+| GitHub | Repository hosting, issues, the roadmap project board, pull requests, Actions, and the artifact and cache stores | GitHub's terms of service | Allowed. Nothing beyond the repository's own content is sent. |
+| Docker Hub | The source of the `pgvector/pgvector` and `greenmail/standalone` images | Docker's terms of service, including anonymous and authenticated pull-rate limits | Allowed for development and integration testing. |
+
+## Components considered and not adopted
+
```
Krzysztof318: Remove considered, only currently used software should be described

THIRD_PARTY_LICENSES.md
```diff
@@ -0,0 +1,200 @@
...
+| MinIO or any S3-compatible object storage | A future raw-content storage migration behind `IEmailContentStore`; not part of the first release | Review server, SDK, client, container image, and trademark terms before introducing the dependency. MinIO's own server licensing is not permissive and needs an explicit owner decision. |
+| Auth0 service and SDKs | An earlier revision recorded Auth0 as the default identity-provider choice. It is not one: issue #143 specifies provider-neutral OAuth 2.1 and OpenID Connect, and no Auth0 package is referenced anywhere. | If a specific provider is ever adopted, review its SDK license, hosted-service terms, pricing, data-processing terms, and trademark rules then. |
+| smtp4dev | Evaluated as the integration-test SMTP target and not selected; GreenMail was chosen instead | Nothing. The decision is recorded with the GreenMail entry above. |
+| System.CommandLine | Named in the architecture draft for the future CLI boundary, which does not exist yet | Pin centrally and record here before first use. |
+
+## Revalidation findings, 2026-07-31
+
```
Krzysztof318: File shouldnt have version history in content

THIRD_PARTY_LICENSES.md
Krzysztof318, line 1: Auth0 to tylko sugestia byla nie cos co ma sie juz znalezc w third party licenses

THIRD_PARTY_LICENSES.md
Krzysztof318: Fixed in 3cd7038 — the `Components considered and not adopted` section is gone entirely, along with the smtp4dev comparison that had crept into the GreenMail row. The rule is now stated in three places so it does not get rediscovered: the register's opening paragraph, its operational rules, and `AGENTS.md`. A row asserts a completed review of software that exists here today; a planned, evaluated, or rejected candidate keeps its reasoning in the specification, ADR, or issue that owns it.

THIRD_PARTY_LICENSES.md
Krzysztof318: Fixed in 3cd7038 — the `Revalidation findings, 2026-07-31` section is gone, as is the "the previous revision recorded / previously missed" phrasing inside individual rows and the revalidation date in the header table. What the register used to say is what version control is for. The two forward-looking items from that section survive as `Open obligations`, because a pending notice-bundle obligation and a prerelease pin are current facts rather than history. The findings themselves are in the PR description.

THIRD_PARTY_LICENSES.md
Krzysztof318: Fixed in 3cd7038 — Auth0 is no longer mentioned anywhere in the file. It went out with the whole section of non-adopted components, so it is not recorded as a dependency, a suggestion, or a deferred review.

THIRD_PARTY_LICENSES.md
```diff
@@ -0,0 +1,200 @@
...
+## Externally sourced source material
+
+| Component and exact version | Use in this repository | License terms verified | Compatibility decision and obligations | Sources reviewed |
+|---|---|---|---|---|
+| Aspire.ProjectTemplates 13.4.6, service-defaults template | The scaffold for health checks, telemetry, resilience, and service discovery, vendored and then rewritten as `src/Host/ServiceDefaultsExtensions.cs`. The file is maintained as MailMcp source and has diverged from the template it started as. | MIT, declared as a license expression in the NuGet package metadata and in the upstream repository | Allowed. MIT permits use and modification under both distribution models, and the resulting file is redistributed as part of `MailMcp.Host`. The emitted template carries no upstream copyright header, so no separate attribution is required; the origin is recorded here so a future notice bundle decides that deliberately rather than by omission. | <https://www.nuget.org/packages/Aspire.ProjectTemplates/13.4.6>, <https://github.com/microsoft/aspire> |
```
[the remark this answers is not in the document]
Krzysztof318: Right, and fixed in 3cd7038. A missing header in generated output is not a grant, and the row was reading it as one. It now records that Microsoft publishes no exception waiving MIT's notice condition for template output, states the obligation as live, and says a distribution must carry Aspire's MIT copyright notice and license text for whatever substantial portion of the template survives in `ServiceDefaultsExtensions.cs`. #113 resolves it by including that notice or by citing an upstream exception; the one option now ruled out is leaving it off on the strength of the missing header. The template row is also linked to `LICENSE.TXT` in the Aspire repository rather than to the repository root.

THIRD_PARTY_LICENSES.md
```diff
@@ -0,0 +1,200 @@
...
+
+## Hosted services
+
+These are called, not linked. None is a redistributable dependency, and none supplies code that ships.
+
+| Service | Use in this repository | Terms that apply | Decision |
+|---|---|---|---|
+| nuget.org | The only package source `NuGet.config` declares, after clearing the machine's inherited source list | NuGet.org terms of use, plus the per-package licenses recorded above | Allowed. Adding a second source is a licensing and supply-chain decision and must be recorded here. |
+| GitHub | Repository hosting, issues, the roadmap project board, pull requests, Actions, and the artifact and cache stores | GitHub's terms of service | Allowed. Nothing beyond the repository's own content is sent. |
+| Docker Hub | The source of the `pgvector/pgvector` and `greenmail/standalone` images | Docker's terms of service, including anonymous and authenticated pull-rate limits | Allowed for development and integration testing. |
```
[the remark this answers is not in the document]
Krzysztof318: Right, and fixed in 3cd7038. Each hosted-service row now carries its official terms URL and the conditions that actually bind, instead of naming a document generically. nuget.org: the Marketplace and NuGet.org Terms of Use, last updated January 2025, recording that Microsoft licenses no package intellectual property and is not a party to a package's own terms — which is precisely why every package license is recorded separately above rather than inferred from the feed — plus the user-data request policy. GitHub: the Terms of Service, the Additional Product Terms that cover Actions, and the Privacy Statement. Docker Hub: the Terms of Service and the Subscription Service Agreement, with the pull-rate limits stated exactly (100 per six hours per IPv4 address or IPv6 /64 subnet unauthenticated, 200 authenticated), classified as an availability condition on the integration suite rather than a licensing one. All five pages were read, not just linked.

THIRD_PARTY_LICENSES.md
```diff
@@ -0,0 +1,200 @@
...
....6.0, Microsoft.Extensions.Diagnostics.ExceptionSummarization 10.6.0, Microsoft.Extensions.Http.Diagnostics 10.6.0, Microsoft.Extensions.Http.Resilience 10.6.0, Microsoft.Extensions.Resilience 10.6.0, Microsoft.Extensions.ServiceDiscovery 10.6.0, Microsoft.Extensions.ServiceDiscovery.Abstractions 10.6.0, Microsoft.Extensions.Telemetry 10.6.0, Microsoft.Extensions.Telemetry.Abstractions 10.6.0 | Provider-neutral chat and embedding abstractions, and the service-defaults stack for HTTP resilience, service discovery, and telemetry enrichment. `Microsoft.Extensions.AI.Abstractions` resolves to 10.8.3 rather than the 10.8.1 pinned centrally, because `ModelContextProtocol.Core` 2.0.0 requires that version and the central pin is a floor. | MIT, declared as a license expression in each package's NuGet metadata | Allowed. Permissive license supports both distribution models with notice preservation. Provider-specific AI packages, model terms, and service terms remain separately reviewable and none is adopted. | <https://github.com/dotnet/extensions>, <https://www.nuget.org/packages/Microsoft.Extensions.AI/10.8.1>, <https://www.nuget.org/packages/Microsoft.Extensions.AI.Abstractions/10.8.3> |
```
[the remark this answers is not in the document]
Krzysztof318: Right, and fixed in 3cd7038. `src/AI` and `tests/AI.UnitTests` resolve `Microsoft.Extensions.Compliance.Abstractions` 10.5.0 while the host graph resolves 10.6.0, and only the latter was recorded. Both versions are now named, with a sentence explaining that this is the second package after `Microsoft.Extensions.AI.Abstractions` where a central pin acts as a floor and nearest-wins raises it per project, and stating which version reaches an artifact in each case. I swept the whole closure for the same shape rather than fixing the one instance: eleven packages resolve at two versions, and the other nine were already covered because the shared-framework row and the test-framework row each carry their own variant.

.agents/skills/check-docs-licenses/SKILL.md
```diff
@@ -31,9 +31,9 @@ For every affected item, verify current official upstream evidence for:
...
+`THIRD_PARTY_LICENSES.md` records what the repository actually pins, bundles, or calls, and is not the project's own `LICENSE` or a generated notice bundle. Put an entry in the section that matches the component's exposure — redistributed, build-time, test-only, orchestration, continuous integration, developer tooling, externally sourced source, or hosted service — because that is what decides whether a release obligation follows. A component the change only mentions, plans, or might adopt later does not belong in the register at all, because a row implies a review that a future adoption would have to redo anyway. The `Components considered and not adopted` section holds those, and states that it carries no review; do not add to it for a dependency the change does not introduce.
```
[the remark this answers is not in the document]
Krzysztof318: Fixed in 3cd7038, though not the way this suggests. The contradiction is real and came from the section it points at, which the repository owner has since removed: rejected and deferred candidates are no longer recorded anywhere. The gate now gives one instruction — register only what the change actually introduces, in any wording, and put that reasoning in the specification, ADR, or issue that owns the future work instead. Removing a component removes its row in the same change.

THIRD_PARTY_LICENSES.md
```diff
@@ -0,0 +1,200 @@
...
+| dorny/paths-filter v4 | The `Detect changes` job of the `CI` workflow, which reads a pull request's changed files through the GitHub REST API and decides whether the build and formatting jobs have anything to do | MIT in the upstream repository, confirmed through the GitHub license API; `v4` is the moving major tag, released as `v4.0.2` | Allowed as continuous-integration tooling. The action runs with a read-only `GITHUB_TOKEN` scoped to `contents: read` and `pull-requests: read`, sends nothing outside the GitHub API, and is not a redistributable dependency. | <https://github.com/dorny/paths-filter/blob/master/LICENSE>, <https://github.com/dorny/paths-filter/releases/tag/v4.0.2> |
```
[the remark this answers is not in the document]
Krzysztof318: Leaving this open, because it is a real point about the wrong file. Pinning `actions/*` and `dorny/paths-filter` to commit SHAs is a change to `.github/workflows/`, which this pull request does not touch and which is a protected path with its own review path; it is a supply-chain policy decision rather than a correction to the register. On the overclaiming: the register does not assert that these steps are immutable. It states that each `v<major>` is a moving tag and that the exact commit a run executes is chosen at run time, which is the accurate description of what is configured today, and it names the reviewed release for `dorny/paths-filter` (`v4.0.2`). Recording a moving tag as moving is not the same as claiming it is fixed. Happy to open a follow-up issue for SHA pinning plus a Dependabot policy to keep the pins current, if the owner wants it tracked — that is his call and not something to fold into a licensing change.

Issue #179, opened 2026-07-31
Serve the health endpoints on a dedicated port with a configurable transport, and let a deployment turn them off


Krzysztof318:
## Context

**No numbered specification under `specs/` backs this issue.** This is first-release hosting infrastructure for the operational probe surface.

Today `ServiceDefaultsExtensions.MapDefaultEndpoints` maps `/health` and `/alive` only when the environment is Development, so a production deployment has no probe surface at all, and a development one serves both endpoints on the same listener that serves `/` and `/mcp`. There is no startup probe at all, and the single registered check is `self`, so readiness and liveness currently answer the same question. Neither the Aspire-inherited environment guard nor the endpoint paths are a project-owned decision: an operator cannot turn the endpoints off, cannot move them off the listener their MCP clients reach, and cannot state which transport carries them.

The environment guard exists in the service-defaults template because an unauthenticated readiness endpoint on a publicly reachable listener reports dependency state to anything that can connect. Deleting the guard without replacing it would publish that surface; keeping it means every container and Kubernetes deployment (#158, #123) has no probe to call. The project-owned answer is exposure control by listener rather than by environment: the probes get a port of their own that an operator publishes to the orchestrator's network and not to the network that serves MCP, and that behaves identically in every environment, so what a developer tests is what runs in production.

Kestrel supports several named endpoints with independent ports and schemes, and a route can be constrained to the connection's local port, which is what keeps `/health` off the application listener and `/mcp` off the probe listener. Official reference: [Configure Kestrel endpoints and HTTPS in ASP.NET Core 10](https://learn.microsoft.com/aspnet/core/fundamentals/servers/kestrel/endpoints?view=aspnetcore-10.0) and [Health checks in ASP.NET Core](https://learn.microsoft.com/aspnet/core/host-and-deploy/health-checks?view=aspnetcore-10.0).

## User stories

**As an operator running MailMcp in Kubernetes or Compose** I want startup, readiness, and liveness probes on a dedicated port that is identical in every environment, **so that** the orchestrator can wait out a slow start, gate traffic, and restart a stuck process on three separate signals without any of them being reachable from wherever MCP clients connect.

**As a security owner** I want the probe surface to be something a deployment turns on, place on its own port, and optionally require TLS on, **so that** an unauthenticated endpoint reporting dependency state is never published by default and never travels clear text on a network where that is unacceptable.

[...]

Issue #156, opened 2026-07-30
Publish releases and nightly builds to GHCR


Krzysztof318:
## Context

**No numbered specification under `specs/` backs this issue.** It is release and nightly-build automation for the first public release.

#116 defines the versioning and release policy, recorded in [ADR 0004](https://github.com/Krzysztof318/MailFathom/blob/main/docs/decisions/0004-versioning-and-release-policy.md); #119 stamps versions into builds; and #123 owns the secure, reproducible, multi-architecture container image. This issue owns **the publication pipeline itself, and GHCR as the registry it publishes to**. #235 adds Docker Hub to the same pipeline as a mirror of the same manifest list.

**This is a split of one piece of work, not a narrowing of it.** ADR 0004's *Channels* section stands unchanged: both channels publish to both registries, as one manifest list pushed twice, under the same digest. That end state is reached when #235 lands. Until it does, a MailFathom release exists on GHCR only, and every document that names a registry says so rather than describing a Docker Hub that is not there yet. The pipeline is built for two registries from the start — the build is performed once and the push is the part that is per-registry — so adding the second one is a login, a tag list, and a verification rather than a second workflow.

**The channels are separated by the version identifier, not by the registry.** A nightly always carries a `-nightly.<n>` prerelease identifier; a release never carries any prerelease identifier at all. That is decidable from the reference alone — by a human reading it and by a schema validating it — and it holds whichever registry the image came from. Every image additionally carries an `io.mailfathom.release-channel` label stating `release` or `nightly` explicitly, so the channel survives a reference that has been re-tagged or reduced to a digest. The separation must be visible in workflow names, image tags, metadata, documentation, and operator examples, so nobody mistakes a development snapshot for a supported release.

**While this repository is private, the package is private too.** The publish run asserts the GHCR package's visibility rather than trusting the default, because a package that quietly turned public would publish an unreleased product's artifacts and every layer inside them.

## User stories

**As an operator** I want an accepted MailFathom release to appear automatically in GHCR under its version, **so that** deploying a stable version is pulling a tag rather than building one.

**As a maintainer** I want a nightly snapshot published every night when `main` has actually moved, and on demand, **so that** the current state of the branch is testable without creating a release and without paying for builds of a commit already published.

**As a security-conscious consumer** I want the release and nightly channels to be unmistakably separate and traceable to their commits, **so that** I know the support and stability level of the image I run.

### The release channel

[...]

Issue #126, opened 2026-07-29
Harden the migration baseline and the add-migration skill for release


Krzysztof318:
## Context

**No numbered specification under `specs/` backs this issue.** It records the part of specification 19 that was deliberately left pre-release, and it carries the `0.1.0 — first public release` milestone because a released instance cannot use the workflow the repository used before it.

The migration workflow was destructive: the repository kept exactly one migration, `Initial`, and `$add-migration` deleted and regenerated it on every model change, recreating the local database by design. That is acceptable only while every database carrying those migrations is a developer's own.

**The freeze belongs before the first published artifact, not at the `v0.1.0` tag.** ADR 0004 originally ended the regeneration policy the moment `v0.1.0` was tagged. The tag is too late: the same ADR establishes a nightly channel built from `main`, and a nightly is installable months before any tag exists. A regenerated baseline changes the migration identifier, so a host started against a database that recorded the previous one fails fast with `32001` and no forward migration can reach it — that installation can only be recreated, destroying whatever it held. A nightly is unsupported, which is a statement about support rather than a licence to strand its data. The local development database that #222 made persistent is the same argument at a smaller scale.

EF Core's intended model is additive: it compares the current model with the checked-in model snapshot, generates a new named migration, and uses the migrations history table to apply only missing migrations.

## User stories

**As an operator upgrading a MailFathom instance**, I want each build to apply only migrations that have not yet run, **so that** schema evolution preserves my stored mail and operational state.

**As a developer**, I want `$add-migration` to require a descriptive name and append one EF Core migration, **so that** the regeneration habit cannot delete a migration another database has already recorded.

**As a reviewer**, I want a model change that arrives without its migration to fail before merge, **so that** the mismatch is caught in the pull request rather than at a host's startup.

## Acceptance

- Freeze `Initial` as the baseline. Its migration ID, generated files, and semantics are immutable; later model changes never delete, rename, regenerate, reorder, or edit it.
- Rewrite `$add-migration` to add one incremental migration. It requires a descriptive migration name, rejects `Initial` and duplicate names, and preserves every existing migration file.
- Retire `scripts/regenerate-migration.sh` with no replacement. No command in the repository deletes a migration. Clearing local data stays `ef-database-reset`, which drops the database and replays the migrations without touching a file.
[...]

Issue #119, opened 2026-07-29
Stamp the application version into every build, report it at runtime, and add the release-preparation skill


Krzysztof318:
## Context

**No numbered specification under `specs/` backs this issue.** It is a release obligation, and it is the mechanical half of the versioning work: #116 decides what a version number means, and this one makes the build, the running process, and the release procedure produce it.

The policy is recorded in [`docs/decisions/0004-versioning-and-release-policy.md`](https://github.com/Krzysztof318/MailFathom/blob/main/docs/decisions/0004-versioning-and-release-policy.md). **That ADR stays `proposed` for now and it does not block this work.** The owner reviews proposed ADRs on a separate track; `proposed` means the reasoning has not been formally accepted, not that the decision is undecided. Implement against it. If the review later moves something, this issue changes with it — which is cheap, because nothing is published and nothing is deployed.

Nothing stamps a version today. `Directory.Build.props` sets no version property, so every assembly in the solution compiles to the MSBuild default of `1.0.0.0` and no build output can be traced back to the commit it came from. The workflows under `.github/workflows/` build and test but publish no named artifact. `deploy/systemd`, `deploy/postgres`, and `deploy/certificates` are empty placeholders, so neither of the two supported deployment shapes — container image and native process — has a naming convention yet.

The runtime consequence is the one that hurts in production. `docs/operations/host-startup-telemetry.md` describes a bootstrap logger that records host startup, and the MCP protocol reports server implementation information during `initialize`; both are places a client or an operator would reasonably expect to learn which build they are talking to, and neither can say anything useful while every assembly claims `1.0.0.0`.

## User stories

**As an operator** I want the running process to report a version that identifies the exact commit it was built from, **so that** a bug report from a deployment I did not build is actionable.

**As an MCP client author** I want the server to report its version during `initialize`, **so that** I can adapt to a tool contract change without probing for it.

**As the project owner** I want the version derived from the repository rather than typed into a file, **so that** releasing does not depend on remembering to bump a literal.

**As the project owner** I want one command to prepare a release and then tell me exactly what to merge and in what order, **so that** the sequence that cannot be automated is not the part I have to remember correctly under time pressure.

### Stamping the version

[...]

PR #180, opened 2026-07-31
Ship MailMcp as a container, a Compose deployment, and a Helm chart

### LINKED ISSUE #158: Provide production Dockerfiles, Docker Compose deployment, and a Helm chart
## Context

**No numbered specification under `specs/` backs this issue.** It is deployment packaging for the first public release.

#123 owns the canonical container image contract, while this issue owns the checked-in Dockerfile and the deployable Docker Compose and Helm assets. #156 publishes stable release images only to Docker Hub and manually triggered nightly builds only to GHCR. Every deployment asset must preserve that distinction: Docker Hub is the safe default; GHCR requires an explicit nightly opt-in and is never presented as a stable release source.

This issue must keep one runtime contract across all three forms. Image identity, configuration keys, secret mounting, ports, health probes, migrations, persistence, TLS boundaries, and shutdown behavior must not acquire a Docker-specific and a Kubernetes-specific meaning.

## User stories

**As a self-hosting operator** I want a supported Docker Compose deployment, **so that** I can start MailFathom and its required infrastructure with documented configuration rather than assembling containers by trial and error.

**As a Kubernetes operator** I want a configurable Helm chart with secure defaults, **so that** I can install the official stable image without maintaining private manifests.

**As a release maintainer** I want the Dockerfile, Compose model, and chart to consume the same image and configuration contract while keeping stable and nightly channels explicit, **so that** users cannot accidentally deploy a nightly as a release.

## Acceptance

- The repository contains production-ready, multi-stage Dockerfile(s) and a `.dockerignore`. The Dockerfile is the checked-in, reviewable image definition used by #123 and #156, builds reproducibly from a clean clone, supports `linux/amd64` and `linux/arm64`, and does not copy source, build tools, credentials, certificates, or unrelated repository artifacts into the runtime image.
- The runtime image runs as a non-root user, supports a read-only root filesystem with explicit writable mounts where required, drops unnecessary Linux capabilities, contains no baked-in secrets, exposes only required ports, handles graceful shutdown, and defines or supports the health/readiness behavior required by #123.
- Docker Compose provides a supported deployment for MailFathom and PostgreSQL with health checks, dependency readiness, restart behavior, named persistent volumes, bounded logs, configurable resource limits, and a clear network boundary.
- Compose uses an immutable stable Docker Hub release image by default and never silently follows GHCR or a nightly tag. An operator may explicitly select a GHCR nightly image for testing, but the configuration requires a deliberate opt-in and visibly warns that it is unsupported development output.
- Compose configuration contains safe, usable non-secret defaults. Credentials, API keys, certificates, and other secrets come from environment indirection or mounted secret files; example files contain placeholders only and are safe to commit.
- Database migration remains an explicit operator action. If Compose supplies a migration service or profile, it is one-shot and manually invoked; starting the host never applies migrations implicitly.
- A Helm chart is added with `Chart.yaml`, documented `values.yaml`, `values.schema.json`, namespaced helpers, and one resource per template file. It renders a Deployment, Service, ConfigMap, references to pre-existing Secrets, health probes, resources, volumes, and an optional Ingress without embedding certificate material.
- The chart defaults to an immutable Docker Hub release image/tag. Selecting a GHCR image requires an explicit nightly-channel value or opt-in, rejects stable-version assumptions that do not hold for nightly builds, and surfaces a warning in chart notes and documentation.
- Chart `version`, `appVersion`, Docker Hub release tags, and the versioning rules in #116/#119 cannot drift silently. GHCR nightly identifiers follow #156 and never become the chart's stable default or `appVersion` release claim.
- Kubernetes defaults satisfy the Restricted Pod Security Standard where the application permits it: non-root execution, no privilege escalation, a read-only root filesystem, dropped capabilities, a seccomp profile, and no automatic service-account token unless explicitly required.
- PostgreSQL ownership is explicit for Kubernetes: the chart either requires an external database by default or offers a separately opt-in dependency. It does not silently install a database, and any chart/image dependency is license-reviewed and recorded in the third-party license register.
- TLS termination, custom domains, ingress class, annotations, certificate Secret names, storage classes, scheduling, affinity, tolerations, topology spread, and resource values are configurable without editing templates. The chart does not attempt to issue or store private certificates itself.
- The same application settings and secret names map predictably between Compose and Helm. All supported inputs are documented, unknown or mutually incompatible values fail validation where the tooling supports it, and no sensitive value is emitted into rendered output or logs.
- CI validates `docker build`, `docker compose config`, Dockerfile/container policy, `helm lint`, schema validation, and deterministic `helm template` output. Smoke verification starts the Compose deployment and installs/upgrades the chart in an ephemeral Kubernetes cluster, checks readiness, and proves clean shutdown and explicit migration behavior.
- Operator documentation provides copy-pasteable Compose and Helm install, upgrade, rollback, migration, secret-provisioning, custom-domain/TLS, health-check, backup, and uninstall commands; states which persistent data survives removal; and clearly distinguishes supported Docker Hub releases from manually produced GHCR nightly builds.
- Base images, Helm dependencies, copied templates, and any new verification tools are reviewed under the repository licensing policy and recorded with required attribution or NOTICE handling.

## Dependencies

- #116 — version and release semantics
- #119 — build and runtime version stamping
- #123 — canonical secure multi-architecture image
- #156 — Docker Hub release and GHCR nightly publication workflows

## Out of scope

Cloud-provider-specific load balancers, managed database provisioning, a Kubernetes operator, a service mesh, automatic certificate issuance, a general-purpose high-availability PostgreSQL stack, or treating GHCR nightly images as a supported release channel.

Krzysztof318:
Closes #158

A production `Dockerfile`, a supported Docker Compose deployment, and a Helm chart — built from one image definition so the three forms cannot drift into three runtime contracts.

Everything lives under `deploy/`: `deploy/docker/` holds the image definition and the ignore-file that bounds its context, `deploy/compose/` the single-machine deployment, `deploy/helm/mailmcp/` the chart.

## One image, and what it deliberately cannot do

`deploy/docker/Dockerfile --target runtime` produces the service, 77 MB, on `aspnet:10.0.10-noble-chiseled-extra`. It carries no migration tool, no SQL, and no credential that could apply one, which makes *the host never applies a migration* a property of what was built rather than a rule someone has to remember.

It runs as UID 1654 on a read-only root filesystem with every capability dropped, writes only to a mounted `/tmp`, exposes 8080, sets `DOTNET_EnableDiagnostics=0` (the dump exposure `secret-provisioning.md` asks deployments to close), and carries OCI provenance labels. The ignore-file is an allow-list, because the build context is a repository root that also holds `.env` files and secrets; it sits beside the Dockerfile as `Dockerfile.dockerignore`, which Docker prefers over the context root's.

Multi-architecture works without QEMU: the app publishes as portable IL on the build platform (`UseAppHost=false`), so the per-architecture stage only pulls a base and copies into it.

## The schema is not here, and neither are the probes

Two things this change previously carried have been returned to the issues that own them, after review:

**The schema step is #126's.** Nothing in `deploy/` applies a schema or renders anything that could — no `migrations` image, no Job, no Compose profile — and `verify-deployment-assets.sh` now fails a definition that reintroduces one. A deployment therefore reaches `DatabaseSchemaOutOfDateException` and stops: the host refuses to serve against a schema no reviewed artifact has applied. Both deployment pages say plainly that establishing it is an operator action until #126 ships.

**The container health probe is #179's.** The image declares no `HEALTHCHECK` — a chiseled image has no shell for one, and adding a probe mode to the application is precisely what #179 decides. What stays is one line: `MapDefaultEndpoints` maps `/health` and `/alive` outside Development. Without it the chart's three probes would ask endpoints that answer 404 in Production, which is the defect #179 opens by naming; this is one line of that issue landing early, and #179 replaces it wholesale.

## Neither deployment defaults to an image

No MailMcp release exists, so nothing here pretends one does. Compose defaults to `mailmcp:local`, a name no registry can serve, and builds this checkout; the chart refuses to render without an explicit repository and an immutable tag or digest, and rejects `latest` and friends outright. GHCR nightlies are reachable only through a deliberate opt-in that fails with the reason when it is missing, and every object a nightly produces is labelled `ghcr-nightly-unsupported`. The nightly identifiers themselves are #156's.

The chart's `appVersion` reads `0.0.0-unreleased` literally. While it does, the image/appVersion drift check is inactive; it activates on its own once #116 and #119 stamp a real one.

## What the deployments serve, and what they leave to you

The chart installs **no database and no Secret**, and templates no certificate material. `database.host` has no default, so it always points at PostgreSQL somebody else operates; `secrets.existingSecret` names a Secret the operator created, mounted read-only at `/etc/mailmcp/secrets`, one file per key. `config.files` renders a ConfigMap mounted as a directory of JSON files. `/etc/mailmcp/secrets` is the same path in both shapes, so a `SecretReference` reads identically either way.

The pod serves plain HTTP with no authentication, no CORS gate, no client certificates, and no rate limiting — the usual Kubernetes arrangement, where an ingress or a mesh in front of the workload owns TLS and client authentication. Each of those is a MailMcp setting rather than a chart value, so enabling one is a ConfigMap entry and nothing else moves; `deployment-kubernetes.md` maps each to its key.

Compose creates an unprivileged `mailmcp` role and installs the `vector` extension during initdb while a superuser is still connected, so the service never needs to be one. The database sits on an `internal` network and publishes no port; MailMcp publishes to loopback, because it terminates no TLS. Kubernetes defaults satisfy the **Restricted** Pod Security Standard, and `values.schema.json` keeps the load-bearing settings from being switched off by accident.

## Verification

```
scripts/verify-deployment-assets.sh   pins, privileges, rendering, determinism, every schema guard
scripts/smoke-deployment.sh compose   the real deployment, from build to the refusal
```

Both were run locally and pass. The Compose smoke proves the image runs unprivileged on a read-only root filesystem, reads its mounted configuration, resolves its mounted secret, reaches the database as the unprivileged role, and then refuses the unrecognized schema — which the startup gate cannot report without having read the migration history. Readiness rejoins the smoke in the change that supplies the schema step.

Full gate: `verify-full.sh` exit 0, 1892 tests, aggregate line coverage **96.90%** (4969/5128), formatting clean over 649 files.

**`scripts/smoke-deployment.sh kubernetes` was not run for this revision**: `kind`, `kubectl`, and `helm` are not installed on this machine. Its Kubernetes half was rewritten along with the rest — install, observe the refusal, compare rendered manifests across a repeated upgrade, uninstall — and is unproven by execution. `helm lint`, `helm template`, and every schema guard did run, through the pinned Helm container image.

## Neither script runs on a pull request

The `Deployment assets` job left `ci.yml` after review, which is now byte-identical to main. It is `.github/workflows/deployment-assets.yml`, `workflow_dispatch` only, exactly like the integration suite, and it is not part of Required CI. It publishes nothing and holds no registry credential; publication is #156.

One consequence: GitHub offers `workflow_dispatch` only for workflows already on the default branch, so this one cannot be started until it merges.

## Worth a look during review

- **`.github/` is a protected path**, so the new workflow file needs your merge specifically.
- **Dependencies.** #116, #119, #123, and #156 are still open, and #126 and #179 now own what this change removed. #187 was opened for publishing the chart alongside the release image as one pipeline.
- **Scope.** This is XL by the sizing rule. You asked for it in one PR after I raised the split; recording that here so the size is not read as an oversight.


deploy/helm/mailmcp/templates/migration-job.yaml
```diff
@@ -0,0 +1,89 @@
...
+              value: {{ .Values.database.port | quote }}
+            - name: PGDATABASE
+              value: {{ .Values.database.name | quote }}
+            {{- /*
+              The migration may connect as a more privileged role than the service: installing the `vector` extension
+              is not something PostgreSQL permits an ordinary role to do. `migrations.user` names that role when the
+              deployment provisions one; leaving it empty reuses the service's.
+            */}}
+            - name: PGUSER
+              value: {{ default .Values.database.user .Values.migrations.user | quote }}
```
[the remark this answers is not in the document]
Krzysztof318: Fixed in 1dd2495. `apply-schema` gained `grant_the_runtime_role_access`, and the Job sets `MAILMCP_RUNTIME_ROLE` to `database.user` only when `migrations.user` differs from it — an owner already has everything, so the same-role case runs nothing.

It grants rather than transfers ownership: handing the tables to the runtime role would leave the migrator unable to alter them on the next migration without membership in that role or superuser, which is the privilege the split exists to avoid. `GRANT USAGE ON SCHEMA public`, DML on all tables, and `USAGE, SELECT` on all sequences cover what exists; two `ALTER DEFAULT PRIVILEGES FOR ROLE CURRENT_USER` statements cover what a later migration creates, so this does not have to be remembered again.

`ci/release-values.yaml` now sets a separate `migrations.user`, so the render covers the branch, and `scripts/smoke-deployment.sh compose` runs the grant against the migrated schema — the `ON ALL` and `ALTER DEFAULT PRIVILEGES` forms are where a syntax or privilege mistake would otherwise only surface in a cluster running the split for real. Both passed.

Dockerfile
```diff
@@ -0,0 +1,196 @@
...
+# The application does not yet report these at run time; stamping them into the assembly is #119, and the release
+# semantics that decide what belongs in `version` are #116.
+ARG IMAGE_VERSION=0.0.0-unversioned
+ARG IMAGE_REVISION=unknown
+ARG IMAGE_CREATED=1970-01-01T00:00:00Z
+
+LABEL org.opencontainers.image.title="MailMcp" \
+      org.opencontainers.image.description="Read-only MCP server over synchronized IMAP mailboxes." \
+      org.opencontainers.image.source="https://github.com/Krzysztof318/MailMcp" \
+      org.opencontainers.image.licenses="Apache-2.0" \
```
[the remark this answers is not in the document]
Krzysztof318: Fixed in 1dd2495. `org.opencontainers.image.licenses` is gone from both image stages and `artifacthub.io/license` from `Chart.yaml`, each replaced by a comment saying the omission is deliberate and that #113 owns the decision — so the next person to notice the missing field adds the LICENSE first rather than the label.

`scripts/verify-deployment-assets.sh` now fails if either reappears, so the two cannot land out of order.

deploy/helm/mailmcp/values.schema.json
```diff
@@ -0,0 +1,293 @@
...
+
+    "terminationGracePeriodSeconds": { "type": "integer", "minimum": 1 },
+
+    "probes": {
+      "type": "object",
+      "required": ["readiness", "liveness", "startup"],
+      "properties": {
+        "readiness": { "$ref": "#/definitions/probe" },
+        "liveness": { "$ref": "#/definitions/probe" },
+        "startup": { "$ref": "#/definitions/probe" }
```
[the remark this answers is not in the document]
Krzysztof318: Fixed in 1dd2495. The shared `probe` definition is split into `readinessProbe` and `processProbe`, and the path is a `const` in each rather than a two-value enum: readiness accepts only `/health`, liveness and startup only `/alive`. The timing fields moved to a `probeTiming` definition both `$ref` so the split cost no duplication.

`probes.liveness.path=/health` is now refused by the schema, and `scripts/verify-deployment-assets.sh` asserts that rejection so the constraint cannot be relaxed silently.

deploy/migrations/apply-schema
```diff
@@ -0,0 +1,126 @@
...
+read_provisioned_file() {
+  local path="$1"
+  local setting_name="$2"
+
+  [[ -r "$path" ]] || fail "${setting_name} names ${path}, which this container cannot read."
+
+  local material
+  material="$(cat -- "$path")"
+
+  printf '%s' "${material%$'\n'}"
```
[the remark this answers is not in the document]
Krzysztof318: Fixed in 1dd2495, in `read_provisioned_file` and in `deploy/compose/postgres/10-create-mailmcp-database.sh`, which had the same shape.

Both now append a sentinel byte the substitution cannot strip and remove it explicitly:

```bash
material="$(cat -- "$path"; printf 'x')"
material="${material%x}"
```

The file then reaches `${material%$'\n'}` byte-for-byte as written, and that one expansion is the only thing that takes a newline off — which is what makes the documented "exactly one" true rather than "all of them". A credential ending in two newlines now authenticates identically here and in MailMcp's own secret resolution.

deploy/helm/mailmcp/templates/_helpers.tpl
```diff
@@ -0,0 +1,171 @@
...
+  {{- fail "secrets.existingSecret is not set. The chart creates no Secret and templates no credential; create one first and name it here." -}}
+{{- end -}}
+
+{{- if .Values.migrations.enabled -}}
+  {{- if not (or .Values.migrations.image.tag .Values.migrations.image.digest) -}}
+    {{- fail "migrations.enabled is true but neither migrations.image.tag nor migrations.image.digest is set. The migration image is versioned with the application it produces the schema for and cannot be inferred." -}}
+  {{- end -}}
+  {{- if not .Values.migrations.image.repository -}}
+    {{- fail "migrations.enabled is true but migrations.image.repository is not set." -}}
+  {{- end -}}
```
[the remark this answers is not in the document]
Krzysztof318: Fixed in 1dd2495. `mailmcp.validate` refuses to render when both images carry a tag and the tags differ, naming both values in the failure and pointing at `migrations.allowVersionMismatch` for the deliberate pairing.

Only tags are compared. Two digests are unrelated strings and say nothing about which versions they are, so a digest pairing states itself and there is nothing to check.

`scripts/verify-deployment-assets.sh` asserts the rejection, `values.yaml` documents why this is refused rather than warned about — a migration from another version is repaired from a backup, not by naming the right image afterwards — and `docs/operations/deployment-kubernetes.md` says the same to an operator.

deploy/compose/compose.yaml
```diff
@@ -0,0 +1,187 @@
...
+
+  mailmcp:
+    # `mailmcp:local` is a name no registry can serve, so an operator who has configured nothing gets the image this
+    # checkout builds and never a tag pulled from somewhere. Point MAILMCP_IMAGE at a published release when one
+    # exists; the release channels and the tags they carry are #156 and #116 and do not exist yet.
+    image: ${MAILMCP_IMAGE:-mailmcp:local}
+    build:
+      context: ../..
+      dockerfile: Dockerfile
+      target: runtime
```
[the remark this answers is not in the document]
Krzysztof318: Fixed in 1dd2495. Both services carrying `image` and `build` now set `pull_policy: ${MAILMCP_PULL_POLICY:-build}`, so the documented `docker compose up -d` builds this checkout instead of resolving `mailmcp:local` through the configured registry.

It is a variable rather than a literal because the pull policy and the image name are one decision: `.env.example` says to change it to `missing` in the same edit that points `MAILMCP_IMAGE` at a published release. The migration service carries the same line with the sharper reason — an image pulled instead of built there would be applying a schema.

deploy/helm/mailmcp/values.schema.json
```diff
@@ -0,0 +1,293 @@
...
+  "required": ["image", "database", "secrets", "config", "service", "probes", "migrations"],
+  "properties": {
+    "nameOverride": { "type": "string" },
+    "fullnameOverride": { "type": "string" },
+    "replicaCount": { "type": "integer", "minimum": 0 },
+
+    "image": {
+      "type": "object",
+      "required": ["channel", "repository", "pullPolicy"],
+      "properties": {
```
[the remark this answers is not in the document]
Krzysztof318: Fixed in 1dd2495. Every fixed-shape object in the schema is now closed with `additionalProperties: false`, including the root, and the genuinely extensible maps are left open by name: `config.files`, `config.extraEnvironment`, `resources`, the annotation and label maps, `nodeSelector`, `affinity`, and `tolerations`.

One addition the closure forced: a `global` property, because Helm injects it when a chart is used as a subchart and a closed root would otherwise reject a parent chart's values. Nothing reads it.

`scripts/verify-deployment-assets.sh` now asserts that `image.pullPolcy` is refused, so a misspelling reports itself rather than leaving the default in place.

deploy/helm/mailmcp/values.schema.json
```diff
@@ -0,0 +1,293 @@
...
+        "mountPath": { "type": "string", "pattern": "^/.+" },
+        "extraEnvironment": {
+          "type": "object",
+          "propertyNames": {
+            "not": {
+              "pattern": "(?i)(password|secret|credential|token|apikey|api_key|privatekey|private_key)"
+            },
+            "description": "An environment variable is readable through /proc by anything in the pod and cannot be erased from process memory. Credentials arrive as mounted files."
+          },
+          "additionalProperties": { "type": ["string", "number", "boolean"] }
```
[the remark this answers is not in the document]
Krzysztof318: Fixed in 1dd2495. `extraEnvironment`'s `propertyNames` became an `allOf` with a second `not`: alongside the credential-word filter, a key matching `^(ConnectionStrings__|Persistence__Password__|ConfigurationSources__|ASPNETCORE_)` is refused outright.

Reserving the prefix rather than validating the value, because the problem is not only the password. The environment block outranks every mounted file, so re-declaring any of these silently replaces what the templates emit — `ConnectionStrings__mailmcp` is merely the case where the override also carries `Password=` into the rendered Deployment. The message points at `database`, `config`, and `secrets` as the places to change instead.

`scripts/verify-deployment-assets.sh` asserts that `config.extraEnvironment.ConnectionStrings__mailmcp=Host=x;Password=y` is refused.

deploy/compose/compose.yaml
```diff
@@ -0,0 +1,187 @@
...
+      - mailmcp-database-password
+    volumes:
+      - postgres-data:/var/lib/postgresql/data
+      # Runs once, on an empty data directory. It creates the unprivileged role and the database MailMcp uses, and
+      # installs the `vector` extension while a superuser is still the one connected.
+      - ./postgres/10-create-mailmcp-database.sh:/docker-entrypoint-initdb.d/10-create-mailmcp-database.sh:ro
+    healthcheck:
+      # Asks for the database MailMcp will use rather than for the server as a whole, so the check does not report
+      # ready while the initialization script is still creating it.
+      test: ["CMD-SHELL", "pg_isready --quiet --username postgres --dbname \"$${MAILMCP_DATABASE}\""]
```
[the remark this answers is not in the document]
Krzysztof318: Fixed in 1dd2495 — the healthcheck is now `pg_isready --quiet --host 127.0.0.1 --port 5432 …`, which only the final TCP-listening server can satisfy, so `depends_on: service_healthy` can no longer fire while `docker-entrypoint-initdb.d` is still running against the temporary socket-only server.

You are right about the second point too, and the comment now says it rather than implying otherwise: the database and user are given for readability, not for enforcement — `pg_isready` reports the server's state and verifies neither.

scripts/smoke-deployment.sh
```diff
@@ -0,0 +1,338 @@
...
+  compose_command logs mailmcp 2>&1 | grep -q "$1"
+}
+
+provision_compose_secrets() {
+  mkdir -p deploy/compose/secrets/mailmcp
+  (
+    umask 077
+    head -c 24 /dev/urandom | base64 | tr -d '\n' > deploy/compose/secrets/postgres-superuser-password
+    head -c 24 /dev/urandom | base64 | tr -d '\n' > deploy/compose/secrets/mailmcp-database-password
+  )
```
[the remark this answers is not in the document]
Krzysztof318: Fixed in 1dd2495. `provision_compose_secrets` is replaced by `provision_smoke_workspace`, which builds a `mktemp` directory holding this run's credentials, an empty configuration directory, and a generated `compose.smoke.yaml` that repoints the secret files and the two bind mounts into it. `deploy/compose/secrets/` and `deploy/compose/config/` are neither read nor written, and the workspace is removed by the same `RETURN` trap that tears the deployment down.

The configuration directory is deliberately empty rather than copied: an operator's own `config/10-mailmcp.json` mounted into this run would make the smoke synchronize somebody's real mailbox.

Verified by running it against a checkout that already had `deploy/compose/secrets/mailmcp-database-password` provisioned — the file's mtime was unchanged afterwards.

scripts/smoke-deployment.sh
```diff
@@ -0,0 +1,338 @@
...
+  pass 'Both images built.'
+}
+
+########################################################################################################################
+# Docker Compose
+########################################################################################################################
+
+compose_command() {
+  (cd deploy/compose && MAILMCP_IMAGE="$runtime_image" MAILMCP_MIGRATIONS_IMAGE="$migrations_image" \
+    docker compose --project-name "$compose_project" "$@")
```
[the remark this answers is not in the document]
Krzysztof318: Fixed in 1dd2495. `compose_command` now exports `MAILMCP_POSTGRES_VOLUME=mailmcp-smoke-postgres-data`, and it does so before the first teardown — the initial `down --volumes` runs through the same function, which is the invocation the comment calls out as the dangerous one.

Verified: a full `scripts/smoke-deployment.sh compose` run creates and destroys `mailmcp-smoke-postgres-data`, and `docker volume ls` afterwards shows the default `mailmcp-postgres-data` name was never created or touched.

deploy/helm/mailmcp/templates/migration-job.yaml
```diff
@@ -0,0 +1,89 @@
...
+        - name: migrate
+          image: {{ include "mailmcp.migrationsImage" . | quote }}
+          imagePullPolicy: {{ .Values.image.pullPolicy }}
+          securityContext:
+            {{- toYaml .Values.containerSecurityContext | nindent 12 }}
+          env:
+            - name: PGHOST
+              value: {{ .Values.database.host | quote }}
+            - name: PGPORT
+              value: {{ .Values.database.port | quote }}
```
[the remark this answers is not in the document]
Krzysztof318: Fixed in 1dd2495. The Job emits `PGSSLMODE` from `migrations.sslMode` and `PGSSLROOTCERT` from `migrations.sslRootCertSecretKey`, resolved against the mounted Secret — libpq reads both directly, so the connection applying privileged DDL now carries a TLS policy of its own.

It is a separate value rather than a reuse of `database.extraConnectionParameters` because those are Npgsql keywords and libpq reads none of them; translating between the two spellings inside a template would be a mapping that silently drifts. Instead `mailmcp.validate` refuses to render when the application configures TLS and `migrations.sslMode` is unset, so the mismatch cannot be reached by omission rather than by choice.

`scripts/verify-deployment-assets.sh` asserts that rejection, and `docs/operations/deployment-kubernetes.md` shows the `verify-full` pairing.

docs/operations/deployment-kubernetes.md
```diff
@@ -0,0 +1,260 @@
...
+
+kubectl --namespace mailmcp create secret generic mailmcp-secrets \
+  --from-literal=mailmcp-database-password='…' \
+  --from-file=imap-primary-password=./imap-primary-password \
+  --from-file=mcp-workstation-key=./mcp-workstation-key
+```
+
+It is mounted read-only at `/etc/mailmcp/secrets`, one file per key, so every credential is a `file:` reference — the
+same path and the same references the Compose deployment uses. A Secrets Store CSI driver needs nothing more, because
+it mounts files too.
```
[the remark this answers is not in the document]
Krzysztof318: Fixed in 1dd2495. The paragraph no longer claims the driver needs nothing more. It now says the chart mounts a Kubernetes `secret` volume and exposes no CSI volume of its own, so the driver's `secretObjects` must **synchronize** into the Secret named by `secrets.existingSecret`, which the chart then mounts like any other.

It also states why mounting the CSI volume directly is not offered: that would need `extraVolumes` and `extraVolumeMounts`, and an arbitrary volume list is how a chart stops being able to say what its pod reads.

deploy/compose/compose.nightly.yaml
```diff
@@ -0,0 +1,34 @@
...
+services:
+  mailmcp:
+    image: ghcr.io/krzysztof318/mailmcp:${MAILMCP_NIGHTLY_TAG:?Name the nightly identifier to deploy. Nightly builds are unsupported development output and carry no release promise.}
+    build: !reset null
+    labels:
+      # Readable from `docker inspect` and from any monitoring that reads labels, so a container started from this
+      # overlay months ago still says what it is.
+      io.mailmcp.release-channel: "ghcr-nightly-unsupported"
+    environment:
+      MAILMCP_NIGHTLY_ACKNOWLEDGED: ${MAILMCP_NIGHTLY_ACKNOWLEDGED:?Set MAILMCP_NIGHTLY_ACKNOWLEDGED=i-understand-this-is-unsupported to deploy a nightly build. It is not a release, its schema and configuration may move without notice, and no support or upgrade path applies to it.}
```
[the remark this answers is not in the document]
Krzysztof318: Kept as documentation rather than enforcement, and the comment now says so plainly rather than implying a check that is not there.

You are right on the mechanics: `${VAR:?…}` tests for a value, Compose interpolation has no equality operator, and the chart's `ne` comparison is therefore strictly stronger. What I disagree with is that closing the gap is worth what it costs. The only way Compose can compare a string is a container that runs one — a validation service the application depends on with `service_completed_successfully` — which buys an ordering dependency, another image in the startup path, and a second mechanism owning a concern the `:?` message already owns.

And it would guard nothing. The acknowledgement is not a security control in either shape: whoever can set `MAILMCP_NIGHTLY_ACKNOWLEDGED=no` can as easily set the exact phrase, and in the chart too. What both are actually for is that nobody reaches a nightly image without reading a sentence saying it is not a release, and `${VAR:?…}` delivers that sentence in full.

So the asymmetry is now stated where someone would otherwise mistake it for a check — in `compose.nightly.yaml`, in the failure message itself, and in `docs/operations/deployment-compose.md`. Left open in case you would still rather have the validation service; that is a call worth your judgement rather than mine.

deploy/docker/Dockerfile.dockerignore
```diff
@@ -0,0 +1,32 @@
...
+# developer's `.env`, a mounted secret, a certificate, and an agent worktree live, and every one of those would be sent
+# to the daemon by a rule that only names what it remembers to exclude. Excluding everything first means a file reaches
+# the build because a line below says it does.
+#
+# The build stage needs the sources and the shared build contract; nothing here decides what reaches the *runtime*
+# image, which copies only the publish output and the generated migration script.
+
+**
+
+!/src
```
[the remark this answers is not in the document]
Krzysztof318: Checked this against Docker's matcher and against a real build, and the premise does not hold: `!/src` does re-include the files below it, and both single-file entries arrive despite their parent directories being excluded.

Docker's `.dockerignore` is not gitignore. `moby/patternmatcher` — which both the classic builder and BuildKit's `fsutil` use — matches a path against a pattern *and* against that path's parent directories (`Matches` compares `parentPathDirs[:len(pattern.dirs)]`), so `src` matches `src/Host/Program.cs` through its parent. Separately, the context walker refuses to `SkipDir` an excluded directory when any exclusion pattern is prefixed by it, which is exactly what keeps `.config/` and `deploy/migrations/` walkable for `!/.config/BannedSymbols.txt` and `!/deploy/migrations/apply-schema`.

The linked "negating matches" section says something narrower than the comment reads it as: in `*.md` / `!README.md`, markdown under subdirectories stays included because `*.md` never matched it in the first place — that is pattern depth, not a rule that a child cannot be restored through an excluded parent.

Proven rather than argued. Building this context with `COPY src/ /ctx/src/`, `COPY .config/BannedSymbols.txt`, and `COPY deploy/migrations/apply-schema`:

```
/ctx/src/Host  /ctx/src/Domain  /ctx/src/Application  … /ctx/.config/BannedSymbols.txt  /ctx/deploy/migrations
--- src file count ---   368
--- bin or obj dirs leaked ---   0
```

All three `COPY` instructions resolved, 368 files reached `src/`, and the trailing `/src/**/bin` and `/src/**/obj` still excluded what they are there to exclude. `scripts/smoke-deployment.sh compose` builds both images from this context on every run, so a context missing its inputs would fail the gate rather than reach review.

Leaving this open rather than resolving it, since it is your finding to withdraw — but no change is warranted.

.github/workflows/ci.yml
```diff
@@ -79,6 +80,22 @@ jobs:
               - 'Directory.Packages.props'
               - 'MailMcp.slnx'
               - 'global.json'
+            # The deployment assets are verified against the application they deploy, not only against themselves:
+            # the smoke run asserts that the host refuses a schema it does not recognize, that its readiness and
+            # liveness endpoints answer, and that it shuts down cleanly. A change under `src/**` can move any of
+            # those without touching a single file under `deploy/`, which is why the whole source tree is here.
+            deployment:
+              - 'Dockerfile'
```
Krzysztof318: jeśli możliwe przenieś DockerFile do deploy/

.github/workflows/ci.yml
```diff
@@ -178,12 +195,70 @@ jobs:
       - name: Verify formatting
         run: dotnet format MailMcp.slnx --verify-no-changes --verbosity diagnostic
 
+  deployment-assets:
```
Krzysztof318: trochę nie rozumiem to ci od razu robi deploy do docker hub i helm hub? raczej nie chce zeby cokolwiek z deploy było cześcią głównego CI przy PR. Deploy ma się uruchamiać tylko dla docker hub i tylko przy podbiciu wersji ale to jeszcze niegotowe bo issue na wersjonowanie jest otwarty

deploy/compose/compose.nightly.yaml
```diff
@@ -0,0 +1,41 @@
+# The GHCR nightly channel. Unsupported.
```
Krzysztof318, line 1: jest na to issue oczekujące możesz dodać do opisu pr

deploy/helm/mailmcp/templates/deployment.yaml
Krzysztof318, line 1: co do helma to domyślnie http bez https, bez auth, bez rate limit bez mtls, bo w k8s najczęściej robi to proxy ale z możliwością vłączenia przez values helm, oczywiście obsługa config map i k8s secrets

Krzysztof318, line 1: a także możliwośc deploy z już istniejącym serwerem/podem postgres bez deploy własnego w helm

deploy/helm/mailmcp/Chart.yaml
Krzysztof318, line 1: tak samo jak obraz dockerowy będzie published na hub.docker tak samo chciałbym helm na oficjalnym chart repository

deploy/migrations/apply-schema
Krzysztof318, line 1: to będzie jako osobny topic, migracje sql, jest na to już issue, nie powinno się znaleźć w tym pr

src/Host/Hosting/ContainerHealthProbe.cs
```diff
@@ -0,0 +1,168 @@
...
+/// endpoints over HTTP from the kubelet and needs nothing inside the container at all, which is why the chart's probes
+/// are HTTP probes; this exists for Docker and Podman, where the platform has no such reach.
+/// </para>
+/// <para>
+/// The probe process is short-lived and deliberately separate from the host it asks about: it composes no
+/// configuration, opens no database connection, resolves no secret, and starts no worker. It therefore reports what a
+/// client of this container would observe rather than what the process believes about itself.
+/// </para>
+/// </remarks>
+internal static class ContainerHealthProbe
```
Krzysztof318: po co to skoro mailmcp wystawia health checks? i jest issue na ich pełne wsparcie startup healthy and alive.

src/Host/ServiceDefaultsExtensions.cs
```diff
@@ -14,8 +14,14 @@ namespace MailMcp.Host;
 /// </summary>
 internal static class ServiceDefaultsExtensions
 {
-    private const string HealthEndpointPath = "/health";
-    private const string AlivenessEndpointPath = "/alive";
+    /// <summary>The readiness path, which reports every registered health check.</summary>
+    internal const string HealthEndpointPath = "/health";
```
Krzysztof318, line 18: out of scope, jest już na to osobne issue

deploy/docker/Dockerfile
Krzysztof318, line 1: jeśli możliwe przenieś do deploy/ jeśli nie zostaw na root

.github/workflows/ci.yml
Krzysztof318: Moved in e50209c. The definition is now `deploy/docker/Dockerfile`, and the ignore-file moved with it as `deploy/docker/Dockerfile.dockerignore` — Docker looks for an ignore-file named after the Dockerfile before it looks at the context root and prefers it, so the file bounding the context travels with the definition that uses it. The build context stays the repository root, so every caller names it: `docker build --target runtime --file deploy/docker/Dockerfile .`

The CI hunk this thread points at is gone entirely: `.github/workflows/ci.yml` is byte-identical to main again. See the reply on the `Deployment assets` thread.

deploy/docker/Dockerfile
Krzysztof318: Moved to `deploy/docker/Dockerfile` in e50209c, with `.dockerignore` alongside it as `Dockerfile.dockerignore`. The repository root now carries neither.

That naming is not a workaround: Docker's build-context documentation defines a Dockerfile-specific ignore-file placed beside the Dockerfile and states it takes precedence over the context root's. So the allow-list that keeps a developer's `.env`, secrets, and certificates out of the build context still applies, and it now lives next to the file it bounds.

.github/workflows/ci.yml
Krzysztof318: It published nothing — no registry credential was available to it, and it never touched Docker Hub or any chart repository. It built the images locally, rendered the chart, and installed it into a kind cluster it created and deleted. Publication is #156 and is not in this pull request.

That said, your intent stands on its own: it cost minutes on every pull request that touched `src/**`. In e50209c it is out of `ci.yml` entirely — that file is byte-identical to main — and lives in `.github/workflows/deployment-assets.yml`, `workflow_dispatch` only, exactly like the integration suite. It is no longer part of Required CI, and it takes a `scope` input so a run can be limited to `compose` or `kubernetes`.

One consequence worth knowing: GitHub only offers `workflow_dispatch` for workflows already on the default branch, so this one cannot be started until it merges.

deploy/compose/compose.nightly.yaml
Krzysztof318: Added. The pull-request description now names #156 for the nightly channel and the tags it will carry, alongside #116, #119, #123, and the new #187.

`compose.nightly.yaml` already said the identifiers themselves are #156's, and it still does; what was missing was the pointer at the top of the description, where a reader looks first.

deploy/helm/mailmcp/templates/deployment.yaml
Krzysztof318: That is what it already does, and nothing had to change for it — but it was not stated anywhere a reader would look, so e50209c adds a *What the pod serves by default* section to `docs/operations/deployment-kubernetes.md`.

The pod serves plain HTTP on 8080 with no authentication, no CORS gate, no client certificates, and no rate limiting. The chart templates no certificate material and asks for no credential to start; TLS belongs to the ingress, which is where `ingress.tls` names a Secret the cluster already holds.

Each of those is a MailMcp setting rather than a chart value, so turning one on is a ConfigMap entry under `config.files` and nothing else moves — which is why the chart has no `auth.enabled` of its own: it would be a second place where one thing is decided. The new section maps each to its key and its documentation:

| To turn on | Configure |
| --- | --- |
| API keys | `McpEndpoint:Authentication`, `McpEndpoint:ApiKeys` |
| An `Origin` gate | `McpEndpoint:Cors` |
| TLS terminated by the pod itself | `McpEndpoint:Https:Endpoints` |
| Client certificates | `McpEndpoint:ClientCertificateProfiles` |
| Rate limits | `McpEndpoint:RateLimiting` |

ConfigMap and Secret support is likewise already the whole configuration path: `config.files` renders a ConfigMap mounted as a directory of JSON files that MailMcp layers beneath the environment block, and `secrets.existingSecret` mounts a Secret the operator created read-only at `/etc/mailmcp/secrets`, one file per key. Every credential is a `file:` reference into that directory, which is also the path the Compose deployment uses, so a `SecretReference` reads identically in both. The chart creates no Secret, and the verification script fails if a template ever renders one.

Krzysztof318: That is the only shape the chart supports, and deliberately: it installs no database at all. `database.host` has no default and the chart refuses to render without one, so it always points at PostgreSQL somebody else operates — an existing server, a pod in the cluster, an operator-managed cluster, or a managed service. `Chart.yaml` declares no dependencies for the same reason, and the verification script asserts the chart renders no database object.

The reasoning is in `values.yaml` and in the Kubernetes page: a store holding every synchronized message needs backup, durability, and an upgrade path a subchart cannot own, and bundling one would make the chart responsible for all three.

What it does need is PostgreSQL with the `vector` extension, and a role that can read and write its own database. Both are stated in `docs/operations/deployment-kubernetes.md` under *What you supply*.

deploy/helm/mailmcp/Chart.yaml
Krzysztof318: Agreed, and it now has an issue: #187 — *Publish the Helm chart alongside the release image as one release pipeline* (`type:workflow`, Track `Release`, milestone 0.1.0).

It is deliberately not only "publish the chart somewhere". The chart is the artifact that *names* the other one — it renders an image reference and its `appVersion` claims which application version it deploys — so a chart published on its own schedule is exactly how a registry ends up holding a chart that points at an image tag the release never produced. #187 therefore owns both halves: an official published location for the chart, and one release run that publishes the image and the chart together, verifies that the chart's default image reference resolves to the image that release actually published, and fails the release when the two disagree.

Whether that location is an OCI registry — Helm 3 pushes a chart as an OCI artifact to the same registry that holds images, so one credential and one version cover both — or a classic `index.yaml` repository, and whether the chart is listed on Artifact Hub, is part of the work rather than something I settled in the issue. Provenance signing and traceability to the commit are in its acceptance list too.

This pull request keeps `Chart.yaml` as it is: `appVersion: 0.0.0-unreleased`, read literally, with the image/appVersion drift check inert until a real version is stamped by #119 under #116's policy.

deploy/migrations/apply-schema
Krzysztof318: Removed in e50209c, and not only the script: the whole schema layer went with it, because leaving any part behind would have left the same topic half-owned. Gone are `deploy/migrations/apply-schema`, the `migrations` stage of the Dockerfile and the SQL-script generation in its build stage, `migration-job.yaml` and the entire `migrations` block of the chart's values and schema, and the Compose `migrate` profile.

The verification script now fails a Dockerfile or a Compose file that reintroduces a schema tool, and fails a chart render that produces a Job, so the boundary is enforced rather than remembered.

One consequence you should see before merging, because it is a real reduction. A deployment now stops at `DatabaseSchemaOutOfDateException`: the host refuses to serve against a schema no artifact has applied. Both smoke halves therefore end there rather than at readiness — which still proves the image runs unprivileged on a read-only root filesystem, reads its mounted configuration, resolves its mounted secret, and reaches the database as the unprivileged role, since the gate cannot report a pending migration without having read the migration history. Readiness rejoins the smoke in the change that supplies the schema step. Both deployment pages now say plainly that establishing the schema is an operator action until #126 ships.

src/Host/Hosting/ContainerHealthProbe.cs
Krzysztof318: Removed in e50209c, with its tests, and the image now declares no `HEALTHCHECK`.

One correction to the premise, because it decides what stays. On main MailMcp does *not* expose health checks where a deployment can use them: `MapDefaultEndpoints` maps `/health` and `/alive` only when the environment is Development, so a production container had nothing to probe. That is the first thing #179 lists as the defect it fixes. So this pull request keeps the one line that removes the environment guard — otherwise the chart's startup, readiness, and liveness probes would be pointed at endpoints that return 404 in the deployment they were written for.

What `ContainerHealthProbe` existed for was narrower and is genuinely #179's: Docker and Podman run a health check as a command *inside* the container, and a chiseled image has no shell and no HTTP client to write one in, so asking the endpoint meant adding a probe mode to the application. Deciding which listener the probes are served on, under which transport, and whether they are served at all is exactly what #179 owns — a probe mode added here would be replaced by that change rather than extended by it. Under Compose the endpoints are asked from outside the container instead, which `docs/operations/deployment-compose.md` now says.

src/Host/ServiceDefaultsExtensions.cs
Krzysztof318: Partly kept, deliberately, and I want to be explicit about which part.

Reverted: nothing else in the file. The constants, the `live` tag, and the endpoint paths are only what removing the environment guard needed.

Kept: `MapDefaultEndpoints` mapping `/health` and `/alive` outside Development. On main they are Development-only, so the chart's three probes and any Compose check would be asking endpoints that answer 404 in Production — the deployment this pull request adds could not be probed at all. #179 owns the full answer: a dedicated listener, a configurable transport, three separate probe kinds, and an off switch. Its acceptance list already states that the Development-only guard in `MapDefaultEndpoints` is removed and replaced by a configured decision, so this is one line of that issue landing early rather than a competing mechanism, and #179 replaces it wholesale.

`ContainerHealthProbe` and its tests are gone — see the reply on that thread. If you would rather this line went too, say so and I will drop the chart's probes with it and let both return together in #179.

PR #339, opened 2026-08-03
[#45] Start a synchronization pass when the mail server says the folder changed

### LINKED ISSUE #45: Spec 11 — IMAP IDLE continuous synchronization
**Roadmap group:** C — continuous synchronization · **Draft stage:** 3
**Specification:** [`specs/11-imap-idle-continuous-sync.md`](https://github.com/Krzysztof318/MailFathom/blob/main/specs/11-imap-idle-continuous-sync.md)
**Depends on:** #38, #43 · **Size:** ~700 lines including tests and documentation

### User stories
**As a mailbox owner** I want new inbox mail to show up promptly through IMAP IDLE instead of on the next poll, **so that** MailFathom reflects my mailbox in near real time.

**As an operator** I want bounded polling to stay available as an explicitly selectable mode and as the fallback, **so that** a server without IDLE still synchronizes predictably.

**As an operator** I want repeated IDLE failures to degrade to polling and later retry push, with the effective mode observable, **so that** I can see which mechanism an account is actually using.

### Acceptance
- An account configured for push uses IDLE on its inbox where supported and polls where not, with the effective mode observable.
- A notification triggers one synchronization pass through the existing synchronizer, not a separate fetch path.
- Repeated IDLE failures degrade to bounded polling, later retry push, and the degradation is recorded.
- `docs/features/imap-synchronization.md` documents mode selection, renewal, degradation, and observability.

Krzysztof318:
Closes #45

Implements specification 11 — IMAP IDLE continuous synchronization.

## What changed

An account configured with `"Mode": "Push"` holds an IMAP `IDLE` session on each of its folders and starts its next synchronization pass the moment the server reports a change, instead of waiting out the interval. Polling stays the default, the fallback, and an explicitly selectable mode.

**Push changes what ends the wait between runs and nothing about what a run does.** The pass a notification starts is the ordinary one — its own read-only session, the same bounded batches, the same checkpoint, the same backward pass. The watching session issues `IDLE` and nothing else, so no second retrieval path exists in which the `\Seen` invariant could hold in one place and lapse in the other. It also does not change the *guaranteed* cadence: the supervisor's wait (interval, or the backoff a failing account is under) is still the longest a folder waits, which keeps reconciliation working through a large folder's window whether or not mail is arriving.

## Design decisions worth reviewing

- **The mode is settled per folder**, against what the server advertises on the connection just established, and re-decided on every reconnect rather than cached. `Domain` gains `MailSynchronizationMode`; `Application` gains `IMailboxNotificationSession(Factory)`, which traffics in domain types and never says "IDLE".
- **Renewal belongs to the caller.** The port takes a `maxWait` and reports `WaitElapsed`; the adapter arms the deadline on `TimeProvider`, so a test drives a renewal without waiting for one. RFC 2177 mandates re-issuing at least every 29 minutes, which is the ceiling on `PushRenewalInterval` (default 20 min).
- **`IDLE` deliberately does not run under the `MailboxDataRetrieval` pipeline.** That pipeline's `AttemptTimeout` is 60 s because a read that stops answering is broken, while a wait that answers nothing for twenty minutes is behaving exactly as designed — running it there would abandon it at every attempt boundary. `MailKitImapConnection.ExecuteUnrepeatedFolderOperationAsync` is the new seam: establishment still runs under `MailboxSessionEstablishment`, so a dropped connection is rebuilt and a rejected credential is still never repeated; only the repetition of the wait itself is given up, and its caller owns that instead.
- **A cancelled wait returns `WaitElapsed` rather than throwing.** The idle state is left in order and the command returns, so the connection stays usable. Throwing would make one folder's notification cost every other watched folder its connection.
- **The connection is a rotation boundary.** A long-lived session has no next connect to pick up a rotation, so each watching session is opened inside a scope pinned to the run's settings snapshot and is recycled between runs — never mid-wait — as soon as a newer one is published. A reload is not distinguishable from a rotation inside it, so every republished snapshot recycles; reconnecting costs one handshake, and the alternative is a revoked credential outliving the reload that replaced it. This makes the rule `docs/operations/secret-rotation.md` already stated hypothetically into implemented behaviour.
- **Degradation expires.** After `MaxConsecutivePushFailures` (default 3) the folder is polled and push is retried after `PushDegradationPeriod` (default 15 min), because a server stops serving `IDLE` for reasons that end.

## Two things a reviewer should push back on if they disagree

- **Push connections sit outside `MaxConcurrentFoldersPerAccount`.** That bound counts folders *synchronizing*; an account in push mode additionally holds one waiting connection per folder, and nothing caps the count. Capping it would mean choosing which folders get push and leaving the rest silently on polling, and the failure a cap would prevent is already handled — a server that refuses the connections produces failures, and degradation turns that into polling folder by folder. Both documentation sections now say this explicitly instead of leaving it to contradict the "one connection per account" default.
- **`MailboxSynchronizationResult` gained a positional parameter** (`MailFolderResolution? Folder`), so the run reports the binding it resolved and the watch can watch the folder the run actually synchronized. It is a `public` type, but no assembly here is published as a NuGet package, so `public` is a cross-project modifier rather than a consumed surface.

## Also in the diff

`MailSynchronizationCoordinatorTests.ExecuteAsync_AccountAddedByAReload_IsSupervisedWithoutARestart` failed once during verification and passed on the next three runs. It is a pre-existing ordering race in the test: the signal it awaits is raised from inside the supervisor, while the assertion is about a line the coordinator writes after `StartSupervisor` returns. Draining before asserting makes it deterministic.

## Verification

- `scripts/verify-full.sh`: pass — 2366 tests, 0 failed; aggregate line coverage 96.46% (required 85%).
- New tests: `MailKitImapNotificationSessionTests` (capability both ways, renewal on `FakeTimeProvider`, notification from three folder events, reconnection after a dropped `IDLE`, cancellation, and the assertion that the push path fetches nothing and sets no flag), `AccountPushNotificationWatchTests` (mode reporting, degradation and its expiry, snapshot recycling, folder removal, disposal), plus `AccountSynchronizationSupervisorTests.RunAsync_WatchedFolderChanges_RunsAnotherPassBeforeTheIntervalElapses` and the options-validation cases.
- `$check-docs-licenses`: Docs pass, Changelog n/a, Licenses n/a — no dependency moved; `Directory.Packages.props` and the lock files are untouched, and MailKit 4.17.0's register row already names IDLE.


docs/operations/configuration-reference.md
```diff
@@ -87,6 +87,9 @@ shape the coordinator loop itself, which are read once at start and marked *rest
 | `MailSynchronization:MaxMimePartCount` | int | `1000` | 1 – 100000 | reload |
 | `MailSynchronization:MaxMimeNestingDepth` | int | `30` | 1 – 1000 | reload |
 | `MailSynchronization:MaxExtractedTextCharacters` | int | `100000` | 1000 – 200000; the ceiling keeps the search vector inside PostgreSQL's limit | reload |
+| `MailSynchronization:PushRenewalInterval` | TimeSpan | `00:20:00` | 1 min – 29 min; the ceiling is what RFC 2177 mandates | reload |
```
Krzysztof318, line 90: Wha is this? I think push mode doesn't have interval only pooling has

PR #383, opened 2026-08-04
[#380] Point the README at the documentation site, and give a page an address that outlives a release

### LINKED ISSUE #380: Point the README at the documentation site, and give the site addresses that outlive a release
## Context

[#122](https://github.com/Krzysztof318/MailFathom/issues/122) published the documentation site and made the root `README.md` name it, but every other documentation link in that file still points at a `blob/main/docs/...` Markdown file on GitHub. A reader who follows one lands in a repository tree — no navigation, no search, no version selector — from the one document that is rendered everywhere the project is listed. The site exists; the front door still sends people past it.

The obstacle is that a site address carries a version. `https://krzysztof318.github.io/MailFathom/v0.3.0/users/README.html` is what the site serves today, and a link written that way is wrong the moment `0.4.0` ships. `latest/` is stable but is the wrong destination: the site deliberately opens on the current release, because `latest` describes the default branch and can document a setting no published version accepts, and a README link should not quietly opt a reader out of that.

So the site needs an address that names a page without naming a version, resolving to whichever version the site currently opens on. Nothing serves one today: `https://krzysztof318.github.io/MailFathom/users/README.html` is a 404.

## Acceptance

- A page has a version-agnostic address at the site root that lands on the default version, and it keeps working when the default version changes.
- Every documentation link in the root `README.md` uses that address, with its heading anchor preserved — docfx and GitHub derive the same slug from a heading, so an anchor carries over unchanged.
- A link to something the site does not publish stays an absolute GitHub URL: the architectural decision records, `docs/README.md`, and the repository-root files.
- `AGENTS.md` states which of the two forms a link in the root `README.md` takes, since the rule there currently says only that every link is absolute.
- A check fails a README link naming a documentation page that does not exist, so the two cannot drift apart silently.

## Out of scope

Rewriting links in any other file. Pages under `docs/` are read inside the repository and keep their relative links, which is what lets a directory be moved without rewriting every reference into it.

Krzysztof318:
Closes #380

Every documentation link in the repository-root `README.md` pointed at a Markdown file in the repository, so the one document rendered wherever the project is listed sent readers past the site it had just published.

## The address problem

A site address carries a version, and neither form a README could use was right. `…/v0.3.0/users/README.html` is wrong the day `0.4.0` ships. `…/latest/users/README.html` is stable but quietly opts a reader out of the release the site deliberately opens on — `latest` documents the default branch and can describe a setting no published version accepts.

So the site now serves both. Composing it mirrors every page of the default version at the site root as a redirect, which gives a page an address naming no version — `…/MailFathom/operations/mcp-endpoint.html` — that lands on whichever version the site currently opens on. The set of stable addresses moves with the pages rather than being maintained beside them: a page a release removes loses its address on the next publish. The API reference is left out; it is a thousand generated pages whose names are type names, nothing links into it by hand, and mirroring it would treble the file count of the site for nothing.

## The links

Forty-six links follow, anchors included — docfx and GitHub derive the same slug from a heading, verified against the deployed site's `container-image.html`, whose `published-images`, `verification`, and `what-a-nightly-build-risks` ids match the README's anchors exactly.

What the site does not publish stays in the repository, deliberately: the architectural decision records, the `docs/` index, and the root files other than `CHANGELOG.md`, which the site does publish.

## Two contracts

Each half of that split fails where nobody reading the file would see it — a site address naming no page is a 404 only a reader meets, and a repository link to a published page quietly delivers a Markdown file instead of the readable form. `every_readme_site_link_names_a_page_that_exists` and `no_readme_link_reaches_a_published_page_through_the_repository` are what make both loud, and `AGENTS.md` now states which of the two forms a link takes rather than only that every link is absolute.

## Verification

- `scripts/verify-full.sh` — pass, 106 workflow contracts.
- `scripts/compose-docs-site.sh` against two version directories — 36 stable addresses covering exactly the published pages, each at the correct relative depth, with the landing page pointing at the default version's front page.
- Both new contracts observed failing: a site link to a page that does not exist and a repository link to `docs/users/usage.md` were each reported by name, and both passed again once removed.


docs/operations/documentation-site.md
```diff
@@ -91,7 +91,23 @@ ADR is an absolute URL for a second reason — a relative path from a source fil
 either.
 
 The repository-root `README.md` is a separate case with a stricter rule, recorded in `AGENTS.md`: every link in it is
-absolute, because it is rendered outside the repository as well.
+absolute, because it is rendered outside the repository as well, and a link to a published page goes to the site
+rather than to the Markdown file behind it.
+
+### The addresses that outlive a release
+
+A page has two addresses. `…/MailFathom/v0.3.0/operations/mcp-endpoint.html` is where it is served, and
```
Krzysztof318: Don't pin 0.3.0 here otherwise every ver update will require update that use v.x.x.x

scripts/compose-docs-site.sh
```diff
@@ -104,26 +106,65 @@ done
...
-    <meta http-equiv="refresh" content="0; url=$default_version/">
+    <link rel="canonical" href="$target">
+    <meta http-equiv="refresh" content="0; url=$target">
   </head>
   <body>
-    <p>The MailFathom documentation is at <a href="$default_version/">$default_version</a>.</p>
+    <p>The MailFathom documentation is at <a href="$target">$label</a>.</p>
   </body>
 </html>
 HTML
```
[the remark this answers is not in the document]
Krzysztof318: Fixed in 9b1f3a8 — you are right that a meta refresh carries no fragment, and that a stub is a page rather than an HTTP redirect, so nothing appended it.

The stub now replaces its own location with the target plus `window.location.search` and `window.location.hash`, and keeps the meta refresh as the no-script fallback. `replace` rather than an assignment, so going back returns to wherever the reader came from instead of to the stub, which would send them forward again.

Checked against every shape the README produces, with the relative resolution made explicit: `/MailFathom/operations/container-image.html#what-a-nightly-build-risks` → `/MailFathom/v0.3.0/operations/container-image.html#what-a-nightly-build-risks`, the same for `features/imap-synchronization.html#transport-security`, an anchorless page unchanged, `CHANGELOG.html#0-2-0` preserved, and `index.html?tabs=compose#top` → `v0.3.0/?tabs=compose#top`, so a query survives as well as a fragment.

docs/operations/documentation-site.md
Krzysztof318: Fixed in 9b1f3a8 — the example now reads `…/MailFathom/v<version>/operations/mcp-endpoint.html`, using the `<version>` placeholder the guides already use, so a release no longer obliges an edit to prose that is not about it.

Worth noting that the paragraph is about the address a reader should *not* use, so the placeholder also makes the point better: what is wrong with that form is precisely that it names a version.

PR #397, opened 2026-08-04
[#375] Make the forwarded-header posture the trust list alone, and say what an unnamed proxy costs

### LINKED ISSUE #375: Decide what an untrusted-peer forwarded-header posture costs, and whether an empty trusted-proxy list may express it
`ReverseProxy` shipped in #315 with one posture: an enabled section names the addresses or CIDR networks it accepts a forwarded scheme and host from, and enabling it without naming one fails startup. This asks for the second posture — reading the two headers from any peer — and for the security property that decision turns on.

**The capability already exists, and only in its explicit form.** `TrustedProxies: [ "0.0.0.0/0", "::/0" ]` passes validation today: `IPNetwork.TryParse` accepts a `/0` prefix, the host bits are zero, and the middleware then believes every peer. So this is not about making something possible that is impossible. It is about whether the implicit form — an enabled section with an empty list — should mean the same thing, and about what either form costs.

**What it costs is one named refusal.** `RefuseATokenThatArrivedWithoutTransportEncryption` in `src/Host/Security/Transport/TransportSecurityExtensions.cs` refuses an OAuth token that did not arrive over TLS, and it decides by reading `HttpContext.Request.IsHttps` — which is `Request.Scheme`, which the forwarded-headers middleware has already rewritten by then. With a trusted proxy named that is correct: behind a terminating proxy the hop to the client genuinely was HTTPS, and the refusal must not fire. With every peer trusted, any client that can reach the listener sends `X-Forwarded-Proto: https` and the refusal stops working, so a reusable credential travels in clear text and is accepted. That is the concrete reason #315 refused the empty list, and it does not go away by being written as `0.0.0.0/0` instead.

**So the decision comes before the switch.** Either the refusal keeps reading the forwarded scheme, and trusting every peer is a posture that knowingly disables it; or the refusal reads the scheme of the connection this process actually accepted, and the two stop being coupled — at the cost that a deployment behind a terminating proxy would then have every OAuth request refused unless something else tells the refusal the client hop was encrypted. Settle that first; the rest is small.

## User stories

- **As an operator whose proxy has no stable address** — a load balancer whose pool changes, an ingress controller I cannot pin to a CIDR — I want a way to say so that does not require me to invent a range, so that a proxied deployment is reachable without a setting I would have to keep correcting.
- **As an operator**, I want whichever form I choose to be visible in my configuration and in a diff, so that "we trust every peer" is something a reviewer can find rather than something that follows from a field being empty.
- **As a security reviewer**, I want the deployment that trusts every peer to be told, at startup, exactly which protection it gave up, so that the posture is a choice somebody made rather than one nobody noticed.

## Acceptance

- The decision above is recorded before the code: whether `RefuseATokenThatArrivedWithoutTransportEncryption` reads the forwarded scheme or the connection's own, and why. If it keeps reading the forwarded scheme, say so where an operator configuring this reads it.
- Whatever form the wide posture takes, it is stated rather than defaulted. An empty `TrustedProxies` on an enabled section either keeps failing startup or becomes the explicit spelling of this posture — not both, and not silently.
- A configured range wide enough to cover peers the deployment cannot have meant is announced at startup, naming what it disabled. `0.0.0.0/0` and `::/0` at minimum; whether a `/8` earns the same line is part of the decision.
- `docs/operations/mcp-endpoint.md` and `docs/operations/configuration-reference.md` state which forms are accepted and what each costs, so the answer is not left to whoever reads `IPNetwork.TryParse`.
- Unit tests cover the wide form parsing and reaching the middleware, the startup announcement, and the refusal's behaviour under whichever scheme the decision settles on.

## Not in scope

- `X-Forwarded-For` and the client address, which #315 left out for reasons that do not change here.
- Deriving `McpEndpoint:OAuth:Resource` from the request, under any trust condition.

Krzysztof318:
Closes #375

`ReverseProxy` carried a switch that could never be flipped on its own: an enabled section without a trusted proxy was refused at startup, so `Enabled` and `TrustedProxies` were one decision written twice. The key is gone. The forwarded-headers middleware is composed and mapped unconditionally, at the position it already held — ahead of the exception handler and every isolation check — and the trust list is what the section decides.

## The default posture, and what it costs

An unconfigured section resolves to `0.0.0.0/0` and `::/0` rather than to nobody.

This is a deliberate reversal of #375's acceptance criterion "whatever form the wide posture takes, it is stated rather than defaulted". It is stated here instead of there: the owner decided the wide posture should be the default after the cost below was put in front of them, so this is not that criterion being met — it is that criterion being overridden.

The cost is one named protection. `RefuseATokenThatArrivedWithoutTransportEncryption` decides by reading `HttpContext.Request.IsHttps`, which is the scheme the forwarded-header policy has already rewritten. With every peer trusted, any client that can reach the listener sends `X-Forwarded-Proto: https` and has a reusable credential accepted over clear text. `X-Forwarded-Host` is believed on the same terms, so the name a `401` challenge and every absolute address carry is a client's to choose. `McpEndpoint:OAuth:Resource` is unaffected — it stays a configured value compared against a token's audience, never a header.

Two tests pin that, because an assertion that a refusal did not fire proves nothing without the control where it does:

- an unconfigured section lets a client's own `X-Forwarded-Proto: https` satisfy the refusal;
- a named proxy makes the same claim from the same peer change nothing, and the token is refused.

The refusal keeps reading the forwarded scheme rather than the connection's own. That was #375's first open question; changing it would refuse every OAuth request behind a terminating proxy unless something else told the refusal the client hop was encrypted, which is a larger design than this change.

## The announcement

`ReverseProxyTrustWarning` now carries two messages, both at startup through `IHostedService.StartAsync` and never per request. The remedy differs — one deployment narrows a range it wrote, the other names a proxy it never named — while what was given up is identical and both lines say so. A merely wide range, a private `/8`, still produces nothing.

## Breaking change

`ReverseProxy:Enabled` no longer binds. Because the section binds strictly, a configuration still setting it stops at startup naming the key rather than starting under a trust nobody chose. Tested for both `true` and `false`. A deployment that had it `true` keeps its behaviour by keeping its list and deleting the key.

## One thing worth a reviewer's eye

- **`MaximumForwardedHops` is kept configurable.** The instruction was "only `TrustedProxies` configurable", which I read as the contrast against the on/off switch. `MaximumForwardedHops` is not a switch, and removing it is a second broken key that would leave a two-proxy chain unconfigurable. Say the word and it goes.
## Implementation note

`TrustedProxies` stays empty rather than pre-populated with the two prefixes. The configuration binder adds to an existing collection instead of replacing it, so a default written into the property would survive alongside whatever an operator configured and every deployment would trust every peer regardless. Trust resolves when the policy is composed; a binding test asserts a configured list leaves no default trust beside it.

`mcp-endpoint.md` also gains the `describes:` markers it was missing. It quotes both warning texts verbatim and documents the whole section, while its marker claimed neither `ReverseProxyOptions.cs` nor either warning — so a future change to a message would have been pointed at `health-endpoints.md`, which says nothing about them.

## Verification

- `scripts/verify-full.sh` — pass on head `c7d01d7`.
- 2713 unit tests, 0 failed.
- `$review-change` — no findings. `$check-docs-licenses` — Docs pass, Changelog n/a, Licenses n/a (no dependency, file, or copied-code change).

A review note asked for the withdrawn-key paragraph to go; it is removed from the reference page and from `ReadFrom`'s remarks in `c7d01d7`. `ReadFrom_TheWithdrawnEnabledKey_FailsRatherThanIgnoringIt` stays, because it asserts current behaviour — an unknown key stops startup — rather than history.

### DIFF
--- a/docs/operations/configuration-reference.md
+++ b/docs/operations/configuration-reference.md
@@ -275,16 +275,22 @@ Two rules are MailFathom's rather than Kestrel's:
 
 ## `ReverseProxy`
 
-Which proxy this process accepts a public scheme and host from, when something in front of it terminates TLS. One
+Which peers this process accepts a public scheme and host from, when something in front of it terminates TLS. One
 section for the whole process rather than one per surface: it runs at the front of the one request pipeline every
 listener shares, so a proxy named here is trusted on each of them.
 [Behind a TLS-terminating reverse proxy](mcp-endpoint.md#behind-a-tls-terminating-reverse-proxy) is the page.
 
+`X-Forwarded-Proto` and `X-Forwarded-Host` are always read; there is no key that switches that off. What the section
+carries is who they are believed from, and **an unconfigured section believes every peer.**
+
 | Key | Type | Default | Constraint | Change |
 | --- | --- | --- | --- | --- |
-| `ReverseProxy:Enabled` | bool | `false` | Off reads neither `X-Forwarded-Proto` nor `X-Forwarded-Host` from anybody | restart |
-| `ReverseProxy:TrustedProxies` | string list | empty | Required non-empty when enabled; refused when configured while it is not. Each entry an IP address or a CIDR network whose host bits are clear — not a DNS name. The framework's loopback default is cleared rather than inherited. `0.0.0.0/0` and `::/0` are accepted and trust every peer, which disables the refusal of an OAuth token that arrived without TLS — see [trust is the connection](mcp-endpoint.md#behind-a-tls-terminating-reverse-proxy) | restart |
+| `ReverseProxy:TrustedProxies` | string list | empty, which trusts `0.0.0.0/0` and `::/0` | Each entry an IP address or a CIDR network whose host bits are clear — not a DNS name. What is named replaces the default rather than adding to it, and the framework's loopback default is cleared rather than inherited. Left empty, or written as `0.0.0.0/0` and `::/0`, it trusts every peer and so disables the refusal of an OAuth token that arrived without TLS — see [what the default costs](mcp-endpoint.md#behind-a-tls-terminating-reverse-proxy) | restart |
 | `ReverseProxy:MaximumForwardedHops` | int | `1` | At least 1; how far right-to-left through each header a value is believed | restart |
+
+> **`ReverseProxy:Enabled` no longer exists.** A configuration still setting it fails at startup naming the key, because
+> every section here is bound strictly. Delete it: the posture it used to select is now the presence or absence of
+> `TrustedProxies`, and a deployment that had it `true` keeps its behaviour by keeping its list.
 
 `X-Forwarded-For` is never read, so the peer MailFathom observes stays the one that opened the connection, and
 `McpEndpoint:OAuth:Resource` stays a configured value rather than anything derived from a header.
--- a/docs/operations/configuration-reference.md
+++ b/docs/operations/configuration-reference.md
@@ -288,10 +288,6 @@ carries is who they are believed from, and **an unconfigured section believes ev
 | `ReverseProxy:TrustedProxies` | string list | empty, which trusts `0.0.0.0/0` and `::/0` | Each entry an IP address or a CIDR network whose host bits are clear — not a DNS name. What is named replaces the default rather than adding to it, and the framework's loopback default is cleared rather than inherited. Left empty, or written as `0.0.0.0/0` and `::/0`, it trusts every peer and so disables the refusal of an OAuth token that arrived without TLS — see [what the default costs](mcp-endpoint.md#behind-a-tls-terminating-reverse-proxy) | restart |
 | `ReverseProxy:MaximumForwardedHops` | int | `1` | At least 1; how far right-to-left through each header a value is believed | restart |
 
-> **`ReverseProxy:Enabled` no longer exists.** A configuration still setting it fails at startup naming the key, because
-> every section here is bound strictly. Delete it: the posture it used to select is now the presence or absence of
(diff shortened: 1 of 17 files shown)


docs/operations/configuration-reference.md
```diff
@@ -275,17 +275,23 @@ Two rules are MailFathom's rather than Kestrel's:
...
+| `ReverseProxy:TrustedProxies` | string list | empty, which trusts `0.0.0.0/0` and `::/0` | Each entry an IP address or a CIDR network whose host bits are clear — not a DNS name. What is named replaces the default rather than adding to it, and the framework's loopback default is cleared rather than inherited. Left empty, or written as `0.0.0.0/0` and `::/0`, it trusts every peer and so disables the refusal of an OAuth token that arrived without TLS — see [what the default costs](mcp-endpoint.md#behind-a-tls-terminating-reverse-proxy) | restart |
 | `ReverseProxy:MaximumForwardedHops` | int | `1` | At least 1; how far right-to-left through each header a value is believed | restart |
 
+> **`ReverseProxy:Enabled` no longer exists.** A configuration still setting it fails at startup naming the key, because
```
Krzysztof318: this note is unnecessary

Krzysztof318: Removed in `c7d01d7` — the note is gone from the reference page, and the same clause is out of `ReadFrom`'s remarks, which had explained strict binding by naming that key instead of by what the rule does.

`ReadFrom_TheWithdrawnEnabledKey_FailsRatherThanIgnoringIt` stays. It asserts current behaviour rather than history — an unknown key stops startup — and it is the one key a real configuration is likely to still carry, so the loud failure is worth pinning.

PR #459, opened 2026-08-05
[#422] Give every network surface its own listener configuration

### LINKED ISSUE #422: Give every network surface its own listener configuration and retire the application listener
Each network surface should state where it is served, in its own section, in the same shape. Today only two of the three do: `AdminEndpoint` and `HealthEndpoints` each bind a socket from their own `BindAddress` and `Port`, while the MCP endpoint has no address of its own and is served on whatever listener the host was started with — `ASPNETCORE_URLS`, `ASPNETCORE_HTTP_PORTS`, `Kestrel:Endpoints`, or `http://localhost:5000` when a deployment names none.

That asymmetry is the source of a class of hazards rather than a cosmetic difference. Kestrel ignores the URL-shaped addresses as soon as any listener is bound in code, so opening the probe or administrative listener silently takes the application listener away — which is why `ApplicationListenerRestatement` and `ConfiguredApplicationListeners` exist at all, to hand those addresses back to the framework. The same mechanism forces the Aspire app model to declare the probe and administrative endpoints with the `tcp` scheme, because an `http` one would join `ASPNETCORE_URLS` and make that port an application listener the host then refuses to start on. Three separate comments in `src/AppHost` exist to explain that workaround.

Giving `McpEndpoint` its own `BindAddress` and `Port` removes the concept of an application listener entirely, and with it the restatement machinery, the `BindsOwnListeners` special case, the asymmetric `ListenerPorts` semantics, and the `tcp`-scheme trap. The two request-serving surfaces then carry one transport shape between them, which is what lets the administrative endpoint receive mutual TLS later without inventing a second way to configure a listener.

## User stories

- As an operator, I state where each surface is served in that surface's own section, so that reading `McpEndpoint`, `AdminEndpoint`, and `HealthEndpoints` tells me every socket the process opens, and no setting outside them can move one.
- As an operator, I serve a surface over HTTP and HTTPS at once, with the clear-text port redirecting by default, so that enabling TLS does not read as an outage to a client nobody has repointed and I can still choose to serve both schemes deliberately.
- As a maintainer, I add mutual TLS to the administrative endpoint by configuring the same profiles the MCP endpoint already accepts, so that a second listener-configuration shape never has to exist.

## Acceptance

- `McpEndpoint` carries `BindAddress` and `Port` and always binds its own listener when enabled. `ApplicationListenerRestatement` and `ConfiguredApplicationListeners` are gone, along with the port-source switch in the composition root and `McpEndpoint`'s empty-when-clear-text `ListenerPorts`.
- `AdminEndpoint`'s transport configuration is the MCP one: the same `Transport` mode, the same `Https` profiles with per-profile domain, socket, minimum TLS version, HTTP versions, and certificate material, and the same clear-text redirect. The deliberate MCP-only settings — CORS and browser-origin validation — stay MCP-only, and the shared shape is what a later mutual-TLS profile drops into.
- Both surfaces accept `Http`, `HttpAndHttps`, and `HttpsOnly`. Under `HttpAndHttps` the clear-text listener redirects to the profiles by default; turning the redirect off is what makes it serve the routes instead. `HttpsOnly` keeps today's guarantee that no clear-text socket stays open behind a profile.
- `HealthEndpoints` keeps its smaller configuration deliberately: one certificate, one domain, no profiles, no redirect, no rate limiting, no credentials. The probes are health checks and their exposure is controlled by which network their port is published on.
- `ASPNETCORE_URLS`, `ASPNETCORE_HTTP_PORTS`, and `ASPNETCORE_HTTPS_PORTS` are refused at startup with a message naming `McpEndpoint:BindAddress` and `McpEndpoint:Port`, rather than being ignored in silence. `Kestrel:Endpoints` is refused on the same terms, as a second mechanism for a decision these sections now own.
- A configuration that enables no surface at all is refused, because Kestrel cannot bind nothing and would otherwise fall back to its own default address and the ASP.NET Core development certificate.
- The Aspire app model publishes every host endpoint as `tcp` and injects the port into the owning section's key, so the `tcp`-scheme workaround stops being a workaround. The developer topology's development-certificate HTTPS endpoint is removed: it exists only because an `https://` address in `ASPNETCORE_URLS` makes Kestrel reach for that certificate, and MailFathom never serves a listener out of one. Local TLS is configured the way a deployment configures it.
- The integration suite reaches the MCP endpoint and the mutual-TLS host through published `tcp` endpoints, reconciled to an HTTP or HTTPS address in the fixture the way the administrative endpoint already is.
- The container image, the Compose file, and the Helm chart set the MCP endpoint's port through its own key instead of `ASPNETCORE_HTTP_PORTS`.
- The root `/` route, which exists only on the application listener being removed, is retired rather than moved: it is an unauthenticated readiness answer that the probe endpoint already gives.

## Breaking change

This breaks a public configuration key. `ASPNETCORE_HTTP_PORTS` is documented as the application listener's address and is set by the image, the Compose file, and the chart; a deployment that sets it itself will be refused at startup rather than quietly ignored. The owner accepted the break when the work was requested. ADR 0004 permits one in a minor version under `0.x`, and the refusal message names the setting that replaces it, so an operator reads what to change rather than discovering a port that moved.

The change stays inside ADR 0002's configuration boundary: it changes which keys exist, not how configuration is read, bound, validated, or reloaded.

### LINKED ISSUE #460: Let surfaces share a listener, and refuse the settings that cannot be shared
#459 gives every surface its own `BindAddress`, `Port`, and `Transport`, and refuses a port another surface already claims. That refusal is the wrong rule for a deployment that wants one socket to serve two of them — the MCP endpoint and the administrative endpoint on `8080`, or all three on one port — which is a reasonable posture for a single-node install behind one ingress, and one an operator should be able to choose.

Sharing a socket is not the same as removing the refusal, though. A port is a property of a listener, and several settings that read as per-surface are really per-socket: a socket is clear text or it is TLS, it redirects or it serves the routes, it asks for a client certificate or it does not, and it presents one set of HTTPS profiles. Two surfaces sharing a port must agree on all of them, and the failure the owner named is exactly that disagreement.

## User stories

- As an operator, I serve the MCP endpoint and the administrative endpoint on one port, so that a single-node install publishes one socket instead of two and my ingress has one backend to route.
- As an operator, I put all three surfaces on one port when I want the probes reachable wherever the endpoint is, and I keep them apart when I do not, because that is my exposure decision rather than one the product takes for me.
- As an operator, a configuration in which two surfaces share a port while disagreeing about what that port does fails at startup naming both sections, rather than starting and then failing to bind, or binding once and silently serving one surface's posture to the other's clients.

## Acceptance

- Two or three surfaces may name the same `Port`. A shared port is accepted rather than refused; a port claimed by exactly one surface behaves as it does today.
- Surfaces sharing a socket must agree on everything that socket decides, and each disagreement is a startup error naming both sections and the setting: `Transport`, `Https:Redirect:Enabled`, `BindAddress`, and — for a shared TLS socket — the HTTPS profiles bound on it, their minimum TLS version, their HTTP versions, and whether a client certificate is asked for. The worked example is the owner's: both endpoints on `8080` while one redirects and the other serves is refused, because one socket cannot do both.
- Kestrel is asked to bind each socket once. Two surfaces sharing a port must not produce two `Listen` calls for it.
- `HealthEndpointIsolation` and `AdminEndpointIsolation` decide from the surfaces bound to the port a connection arrived on, not from one surface's port set. Today each asks `ports.Contains(localPort) == IsPathMine(path)`, which on a shared port serves that surface's paths and refuses every other surface's — the MCP route would answer `404` on a port it is served on.
- `McpEndpoint` and `AdminEndpoint` default to `8080` for the clear-text socket and `8443` for an HTTPS profile, so enabling both with no ports configured shares one socket and is valid, rather than being a collision an operator has to resolve before the deployment starts.
- Sharing is a posture an operator selects, and the documentation says what it costs: the probes and the administrative surface are exposed by which network their port is published on, so putting them on the endpoint's port publishes them wherever the endpoint is. That is the operator's call and the reason for the startup report that names it.

## Why this belongs before #459 merges

#459 documents and enforces "a port one surface binds is refused to the others", and moves `AdminEndpoint:Port` to `8090`. Both are reversed here. Shipping them a release apart would move a public configuration key twice and document a behaviour that lasts days, so this should land on `agent/unify-endpoint-listener-configuration` and go out as part of #459 rather than after it.

Krzysztof318:
Closes #422
Closes #460

## What moved

The MCP endpoint had no address of its own. It was served on whatever listener the host was started with — `ASPNETCORE_URLS`, `ASPNETCORE_HTTP_PORTS`, `Kestrel:Endpoints`, or `http://localhost:5000` — while the administrative endpoint and the probes each bound a socket from their own section. Kestrel drops the URL-shaped addresses as soon as any listener is bound in code, so opening either of those took the application listener away from a deployment that had stated it. `ApplicationListenerRestatement` existed to hand those addresses back, and the Aspire app model had to declare every endpoint with the `tcp` scheme so Aspire would not issue `ASPNETCORE_URLS` and trip the host's own refusal — a workaround documented in three separate comments.

`McpEndpoint` now carries `BindAddress`, `Port`, and `Transport`, the same four settings `AdminEndpoint` states. The process opens exactly the sockets its endpoint sections describe, the restatement machinery and its reader are gone, and the `tcp`-scheme workaround has stopped being one.

## `Transport`

| Value | The socket at `BindAddress`:`Port` | `Https:Endpoints` |
| --- | --- | --- |
| `Http` (default) | Serves the routes | Must be empty |
| `HttpAndHttps` | Redirects to the profiles, or serves the routes when `Https:Redirect:Enabled` is `false` | Bind |
| `HttpsOnly` | Not opened at all | Bind |

The redirect therefore loses its own `BindAddress` and `Port` — it describes the surface's one clear-text socket — and with them the two divergent defaults the surfaces carried, `8080` for MCP and `8091` for the administrative endpoint. One `EndpointTransport` now answers the same question for all three surfaces, replacing `HealthEndpointTransport`; the probes keep their smaller set of TLS settings deliberately, because they carry one certificate rather than profiles and are reached without a credential.

The listener rules the two request-serving surfaces share live in `TransportListenerConfiguration` rather than being restated per section, and one `TransportListenerBinder` binds both. That is what makes mutual TLS on the administrative endpoint a profile on a listener already shaped to carry one rather than a second way to configure a socket. The profiles themselves are deliberately not added here: a setting nothing reads is the shape this codebase refuses everywhere else.

## What is now refused rather than ignored

- `ASPNETCORE_URLS`, `ASPNETCORE_HTTP_PORTS`, `ASPNETCORE_HTTPS_PORTS`, and `Kestrel:Endpoints`, each with a message naming the key that replaces it. Ignoring them is the silent failure this closes: an operator states a port and the surface answers somewhere else.
- A configuration enabling no surface at all, which would otherwise leave Kestrel binding its own default address plus a TLS one wherever a development certificate happens to be installed.
- Client certificate profiles under a transport that terminates no TLS, since no certificate could ever be presented to judge against them.

The root `/` route is retired with the listener it lived on. It was an unauthenticated readiness answer the probe endpoint already gives.

## The development certificate

The developer topology's `WithHttpsEndpoint(8443)` is removed. It worked only because an `https://` address in `ASPNETCORE_URLS` makes Kestrel reach for the ASP.NET Core development certificate — the mechanism being retired, and one MailFathom already refuses for its own listeners. A developer wanting local TLS now configures `McpEndpoint:Https` the way a deployment does, which is also the shape they ship. The integration suite never used it: the mutual-TLS host serves TLS from material the suite generates, and the ordinary integration host is plain HTTP.

## Breaking change

`ASPNETCORE_HTTP_PORTS` is a documented public configuration key and is set by the image, the Compose file, and the chart. All three now set `McpEndpoint__Port`; a deployment that set one of the refused variables itself fails startup rather than serving somewhere else. The owner accepted the break when the work was requested. ADR 0004 permits one in a minor version under `0.x`, and each refusal names the setting that replaces it. The change stays inside ADR 0002's configuration boundary: it changes which keys exist, not how configuration is read, bound, validated, or reloaded.

## Verification

- `scripts/verify-fast.sh` and `scripts/verify-full.sh` both green on the rebased branch.
- 2705 unit tests pass, 28 of them new across `TransportListenerConfigurationTests` and `ExternalListenerConfigurationTests`.
- Aggregate line coverage 94.62% (7109/7513), against the 85% gate.
- `helm lint` and `helm template` clean against both `ci/*-values.yaml`; the rendered deployment sets `McpEndpoint__Port` beside the container port it restates.
- Net −587 lines across 51 files.

The integration suite is not run locally by policy; the Aspire app model and `MailFathomOrchestrationFixture` are updated in this change, and every host endpoint is now a `tcp` endpoint injecting its port into the section that owns it.

---

## Also closes #460 — surfaces may share a listener

The rule this pull request first introduced, that a port one surface binds is refused to the others, turned out to be the wrong one for a single-node deployment behind one ingress. Two surfaces, or all three, may now name one port.

Sharing is not the same as dropping the refusal. Several settings that read as per-surface are really per-socket, so what replaces it is an agreement check: each surface declares the sockets it wants, `ListenerComposition` groups those declarations by socket, refuses the surfaces that disagree about one, and merges the profiles of those that do not.

| Disagreement on a shared socket | Refused because |
| --- | --- |
| One surface's `Transport` terminates TLS there, the other's does not | One socket serves one scheme |
| One redirects the clear-text socket to its profiles, the other serves its routes on it | One socket cannot do both — the owner's worked example |
| The probes present one certificate where an endpoint selects one by server name | A socket answers a handshake one way |
| Only one surface asks the client for a certificate | Settled while the connection is established, so one answer per socket |
| Two profiles publish one domain on the socket | Composition order would decide which surface a client reached |

Two profiles publishing *different* domains on one TLS socket is accepted and is what sharing one is for; a shared TLS socket resolves a server name against each contributing endpoint's own store in turn, so neither one's material enters the other's section. A wildcard address beside a specific one on a port is still refused, and two specific addresses on a port remain two sockets.

Kestrel is asked to bind each socket once, which is the half that would otherwise fail at runtime rather than at startup.

**The two isolation middlewares become one.** Each asked whether its own port set contained the port a connection arrived on, which on a shared port served that surface's paths and refused every other surface's — `/mcp` would have answered `404` on a port it was served on. `SurfaceIsolation` reads the composed surface set for the port instead, so a path is served exactly when that port carries the surface that owns it.

**Defaults.** `McpEndpoint` and `AdminEndpoint` both default to `8080` clear text and `8443` for a profile, so enabling them together shares one socket rather than colliding. `AdminEndpoint:Port` therefore moves from `8090`, which never shipped — it is introduced and superseded within this pull request.

What sharing costs is exposure: the probes answer without a credential, and administering the service is a different authority from reading the mailbox. It stays something an operator selects, and the reference says so where the ports are described.

Verification after the second commit: both gates green, 2679 unit tests pass with 27 more across `ListenerCompositionTests` and `SurfaceIsolationTests`, aggregate line coverage 94.62%.


Krzysztof318:
Hold this one before merging. #460 changes two things this pull request introduces: the refusal of a port another surface claims, and `AdminEndpoint:Port` moving to `8090`. Both are reversed there — surfaces may share a socket, the settings that socket decides have to agree across the surfaces on it, and both request-serving endpoints default to `8080` clear text and `8443` TLS so that enabling them together shares one socket rather than colliding.

Shipping the two a release apart would move a public configuration key twice and document a behaviour that lasts days, so #460 lands on this branch and goes out as part of this pull request.

Krzysztof318:
P1 on `src/AppHost/Program.cs` — fixed in 3095e65, and it was a real defect rather than a style point. The mutual-TLS host configured HTTPS profiles and a `Required` client-certificate profile while inheriting the clear-text `Transport` default, which this branch refuses precisely so that a profile nothing serves cannot read as TLS an operator enabled. It set `McpEndpoint__Transport` to `HttpsOnly` and the suite's own host would otherwise have failed startup on the rule this branch added.

The branch has also grown a second commit since the review: #460 makes surfaces able to share one socket, replaces the port-collision refusal with an agreement check across the surfaces on a socket, folds the two isolation middlewares into one that reads the composed surface set for a port, and moves both request-serving surfaces onto `8080`/`8443` by default. The pull request body has the whole of it.

deploy/compose/compose.yaml
```diff
@@ -93,7 +93,9 @@ services:
         condition: service_healthy
     environment:
       ASPNETCORE_ENVIRONMENT: Production
-      ASPNETCORE_HTTP_PORTS: "8080"
+      # The MCP endpoint's own port. MailFathom refuses ASPNETCORE_HTTP_PORTS: each surface states where it is served
```
Krzysztof318, line 96: And alo port expose?

deploy/docker/Dockerfile
```diff
@@ -144,18 +144,20 @@ LABEL org.opencontainers.image.licenses="Apache-2.0" \
...
 
-# 8080 is what the .NET base images already listen on, restated here so the number is visible beside the EXPOSE that
-# publishes it. The container speaks plain HTTP: TLS terminates at the ingress or the reverse proxy in front of it,
-# which is also the only place a certificate needs to exist.
-ENV ASPNETCORE_HTTP_PORTS=8080 \
+# The MCP endpoint's own port, stated in the section that owns it. MailFathom refuses ASPNETCORE_HTTP_PORTS and every
+# other URL-shaped address, because each surface states where it is served in its own configuration section. 8080 is
+# restated here so the number is visible beside the EXPOSE that publishes it. The container speaks plain HTTP: TLS
+# terminates at the ingress or the reverse proxy in front of it, which is also the only place a certificate needs to
+# exist.
```
Krzysztof318, line 151: Same whst with api port expose?

deploy/compose/compose.yaml
Krzysztof318: No port for it, and 278cace makes both files say why rather than leaving the silence to be read as an oversight.

The administrative endpoint now defaults to `8080` — the MCP endpoint's own — so enabling it without stating a port serves it on the socket that is already published rather than on one nobody exposed. That is a decision worth taking rather than inheriting, since administering the service is a different authority from reading the mailbox, so both files now point at giving it an `AdminEndpoint__Port` of its own and publishing that instead. Adding a second `EXPOSE` or a second mapping for a surface this image leaves disabled would have described a socket nothing binds.

deploy/docker/Dockerfile
Krzysztof318: No port for it, and 278cace makes both files say why rather than leaving the silence to be read as an oversight.

The administrative endpoint now defaults to `8080` — the MCP endpoint's own — so enabling it without stating a port serves it on the socket that is already published rather than on one nobody exposed. That is a decision worth taking rather than inheriting, since administering the service is a different authority from reading the mailbox, so both files now point at giving it an `AdminEndpoint__Port` of its own and publishing that instead. Adding a second `EXPOSE` or a second mapping for a surface this image leaves disabled would have described a socket nothing binds.

src/AppHost/Program.cs
```diff
@@ -316,7 +324,7 @@
     // opens one number rather than two.
     var mutualTlsHost = builder
         .AddProject<Projects.Host>(OrchestrationContract.MutualTlsHostResourceName, launchProfileName: null)
-        .WithHttpsEndpoint(name: OrchestrationContract.MutualTlsHostHttpsEndpointName)
+        .WithEndpoint(name: OrchestrationContract.MutualTlsHostHttpsEndpointName, scheme: "tcp")
```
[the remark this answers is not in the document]
Krzysztof318: Fixed in 3095e65 — the mutual-TLS host now states `McpEndpoint__Transport=HttpsOnly` beside its profiles. The finding was right: it configured HTTPS profiles and a `Required` client-certificate profile while inheriting the clear-text default, which this branch refuses precisely so that a profile nothing serves cannot read as TLS an operator enabled, and the suite's own host would have failed startup on the rule this branch added.

PR #466, opened 2026-08-05
[#463] Serve the probes on the MCP socket in the integration topology

### LINKED ISSUE #463: Serve the probes on the MCP socket in the integration topology, and stop asserting the retired root route
The integration suite fails on `main`. `ComposedMcpEndpointSecurityTests.ReadinessResponse_WithNoCredential_StillAnswersBecauseOnlyTheMcpRouteRequiresOne` asks the composed host for `/` and expects `200`, and #422 retired that route deliberately — it was an unauthenticated readiness answer the probe endpoint already gives. The test now reads `404` and the suite is red, which the pull-request checks cannot see because the suite runs only on manual dispatch.

The claim the test carries is still worth making: that the credential requirement is carried by the MCP route rather than by a fallback policy, so a probe that has nothing to present is still answered. What it needs is a path that exists. The composed topology serves the probes on a listener of their own, so reaching them from the MCP client proves nothing about that listener's pipeline — and #460 made a second thing possible that nothing exercises end to end: two surfaces sharing one socket, with the isolation middleware serving the union of what they answer and refusing everything else.

Putting the probes on the MCP socket in the integration topology answers both. The probe path is then reachable from the same client, unauthenticated, on the socket the MCP route requires a credential on; and the administrative surface, which stays on a port of its own there, is refused on it.

## User stories

- As an operator sharing one port between the MCP surface and the probes, I want the suite to prove that a real Kestrel serves both on it, so that the composition rules the host validates are known to describe what it then does.
- As a maintainer, I want the integration suite green on `main`, so that a red run is a signal rather than a known failure everyone has learned to skip.
- As an orchestrator, I want a liveness probe answered without a credential on whichever socket it is served from, so that a deployment that shares a port does not fall out of rotation.

## Acceptance

- The integration topology serves the probes on the MCP endpoint's socket, with both sections stating the same bind address so the two cannot drift into a startup refusal.
- One test proves the shared socket serves the union: the liveness probe answers `200` with no credential, and an administrative route on the same socket is refused.
- The retired root route is asserted nowhere.
- The stale reference to the deleted `AdminEndpointIsolation` in the administrative endpoint tests' remarks names `SurfaceIsolation` instead.
- The suite passes on manual dispatch.

Follows #422 and #460.

Krzysztof318:
Closes #463.

The integration suite was red on `main` after #459 merged. `ComposedMcpEndpointSecurityTests.ReadinessResponse_WithNoCredential_StillAnswersBecauseOnlyTheMcpRouteRequiresOne` asked the composed host for `/` and expected `200`, and #422 retired that route deliberately — it was an unauthenticated readiness answer the probe endpoint already gives. The pull-request checks could not see it: the suite runs only on manual dispatch.

## What changed

The claim the test carried is still worth making — the credential requirement is on the MCP route rather than on a fallback policy, so a probe with nothing to present is answered — and #460 made a second one possible that nothing exercised end to end: two surfaces on one socket, where the isolation serves the union of what they answer and refuses everything else.

- The integration topology serves the probes on the MCP endpoint's own socket, by handing the probe section the port that endpoint allocated rather than allocating a second one. A second endpoint would be a second allocation and nothing would make the two the same number.
- Both sections state the same bind address, from one constant. They describe one socket, and an address written in one place and defaulted in the other is two sockets the host refuses to open — a startup refusal that would reach the suite as a host that never started.
- The administrative endpoint keeps a listener of its own, so the suite carries both arrangements at once.
- One test replaces the old one and proves both halves: the liveness probe answers `200` with no credential, and an administrative route on the same socket is refused `404` rather than `401`, which is what says the isolation ran ahead of the credential check.
- Liveness rather than readiness, because it consults process-local state — the answer is then a fact about the pipeline instead of about what the database was doing at that moment.
- The developer topology is unchanged in behavior. Its probes stay on a loopback socket of their own, now stated where that socket is declared rather than on the shared builder, beside the reason it is separate.
- The administrative endpoint tests' remarks named `AdminEndpointIsolation`, which #422 deleted; they name `SurfaceIsolation` now.

## Verification

`scripts/verify-full.sh` green: 108 PASS, 0 FAIL, 2693 unit tests, aggregate line coverage 94.62%.

The integration suite is what this change is about and it runs on manual dispatch only, so it is dispatched against this branch rather than asserted here; the result is reported in a comment below.


Krzysztof318:
Integration suite dispatched against this branch: [run 30999505809](https://github.com/Krzysztof318/MailFathom/actions/runs/30999505809) — **59 passed, 0 failed**. `SharedSocket_WithNoCredential_AnswersTheProbeAndServesNoAdministrativeRoute` passed, which is the assertion that only holds when the probes really are served on the MCP endpoint's socket: the client reaches `/alive` at the MCP base address, and a listener of their own would answer that `404`.

For reference, the failing run this fixes is [30997285143](https://github.com/Krzysztof318/MailFathom/actions/runs/30997285143) on `main` — 58 passed, 1 failed.

Issue #495, opened 2026-08-06
Distribute mfctl to Debian and Ubuntu through apt


Krzysztof318:
No `specs/` file backs this. On Debian and Ubuntu, getting `mfctl` today means downloading
`mfctl-<version>-linux-x64` or `mfctl-<version>-linux-arm64` from a release, verifying it against
`mfctl-<version>.sha256`, renaming it, marking it executable, and moving it onto `PATH`. Nothing about that is
wrong, and nothing about it is how a Linux operator expects to obtain a command on the platform that is the
service's only supported runtime.

There are two distinct pieces of work behind the word `apt`, and they cost very different things.

The first is a `.deb` attached to a release, installed with `apt install ./mfctl_<version>_amd64.deb`. It needs
no hosting, no key, and no new dependency: `dpkg-deb` is already on the runner, and the package is one binary
under `/usr/bin`, a `control` file, and the license. It gives an operator a tracked installation and an
uninstall path, and it does not give them `apt install mfctl`.

The second is an APT repository, which is what that command actually requires, and it is a different kind of
commitment. An APT repository is stateful: `dists/` and `pool/` must keep every version that was ever
published, so something has to hold that history across releases and serve it at a stable address. Its
`Release` file must be signed, so there is a GPG key with an owner, a custody story, and a rotation story, and
its public half has to be published somewhere an operator can fetch into `/etc/apt/keyrings/`. GitHub Pages is
the obvious host and is already occupied by the documentation site that `scripts/compose-docs-site.sh` builds,
so sharing it is a decision about that site rather than an extra directory; #485 is what a slow Pages queue
already costs a publish. And re-releasing a version acquires a third place where that version lives, beside
the two registries the immutability guard already checks.

The alternatives are not obviously cheaper. A Launchpad PPA builds from source, which is a poor fit for a
self-contained trimmed .NET publish, and covers Ubuntu only. A hosted package service moves the key custody
off this repository at the cost of a supply-chain and licensing review of the service itself, and a row in
`THIRD_PARTY_LICENSES.md` for something an operator's `apt` trusts.

None of that decides anything yet. This issue exists so the question is recorded with its cost visible rather
than rediscovered, and it carries no milestone deliberately: it is outside the current release until the owner
decides which of the two shapes — or neither — the project wants to own.

## User stories

**As a** Debian or Ubuntu operator **I want** to install `mfctl` with my system package manager, **so that** the
[...]

Issue #492, opened 2026-08-06
Remove Windows code signing from the release pipeline


Krzysztof318:
No `specs/` file backs this. It follows from an answer the project received rather than from a design
decision: SignPath Foundation declined MailFathom's application to its free program for open-source
projects, because the repository is not established enough to meet the program's eligibility bar. There
is therefore no certificate to sign with, no date by which one arrives, and no second free program this
pipeline was written against.

#363 built the signing pipeline, #389 switched it off in `Release` while keeping every part of it in the
tree, and #390 was the issue that would have turned it back on. What #389 preserved was preserved against
a certificate that was expected; a refusal removes the thing it was waiting for, so the whole arrangement
goes rather than staying switched off indefinitely. Machinery nobody can enable is machinery that ages,
misleads a reader about what a release carries, and keeps an unlicensed action and an unresolved
licensing obligation in the register for a capability the project does not have.

Removing it changes nothing about what a release publishes today: no release has ever carried an
Authenticode signature, and the checksum file is already what verifies every `mfctl` download.

## User stories

- As an operator reading the documentation, I want it to say plainly and in one voice that no `mfctl`
  binary is signed and that the checksum file is what verifies a download, so that I am not routed to a
  page describing a signature that is never coming.
- As the owner, I want the release pipeline to contain no job, secret reference, or repository-variable
  contract for a service the project has no access to, so that `Release` is exactly the set of steps that
  can actually run.
- As somebody auditing the third-party register, I want no row and no open obligation for a dependency
  the repository does not use, so that the register keeps stating what is in use today rather than what
  was once prepared.

## Acceptance

- `.github/workflows/sign-cli-binaries.yml` is deleted, and with it the repository's only Windows runner,
  its only commit-pinned action, and the `SIGNPATH_*` variable and secret contract. None of those
  variables or the secret was ever created in repository settings, so nothing is left to revoke.
- `Release` carries no `Sign the CLI binaries` job. `Publish the GitHub release` keeps attaching what
  `CLI binaries` produced, under the same guard, so the failure behaviour of the command assets is
  unchanged.
- `Build the CLI binaries` publishes no `artifact-id` output. Its only consumer was the signing job, and
  the id existed so SignPath could fetch the artifact through the GitHub API.
- The release notes state that no binary carries a code signature and that the checksum file is what
  verifies a download, with no SignPath attribution and no promise of a future one.
[...]

PR #498, opened 2026-08-06
[#494] Offer mfctl through the Windows Package Manager on every release

### LINKED ISSUE #494: Submit the mfctl manifest to the Windows Package Manager when a release publishes
No `specs/` file backs this. `mfctl` reaches a Windows operator as a file: a release attaches
`mfctl-<version>-win-x64.exe` and `mfctl-<version>-win-arm64.exe`, and installing one means downloading it,
verifying it against `mfctl-<version>.sha256`, renaming it, and putting it somewhere on `PATH`. The Windows
Package Manager removes all four steps, and the binaries already have the shape it takes: one self-contained
file per architecture, at a version-specific URL that never moves, with a published checksum.

The community repository takes a three-file manifest set at schema `1.12.0` — version, defaultLocale, and
installer — under `manifests/m/MailFathom/mfctl/<version>/`. A bare `.exe` is `InstallerType: portable` with
`Commands: [mfctl]`, which is what makes the command reachable under its own name rather than under the asset's
file name, and `jqlang.jq` is the same shape against the same kind of release asset.

Nothing signs a `mfctl` binary and #492 settles that nothing will, so what winget verifies is the
`InstallerSha256` the manifest carries and nothing beyond it. That is the guarantee the checksum file already
gives a manual download; the manifest states it once for the operator instead of asking them to run
`Get-FileHash` and compare by eye.

Submitting the manifest needs a credential the pipeline does not have. A pull request to a repository the owner
does not own can only be opened by a classic token with the `public_repo` scope — a fine-grained token creates
the commit and then fails on the pull request, which is a documented GitHub limitation rather than a
configuration mistake. That scope is account-wide, which is the same objection that keeps a classic `project`
token out of this repository's secrets, and it is accepted here as a decision rather than passed over: the
alternative is a manual submission per release, and the owner has chosen that the pipeline carries it.

## User stories

**As a** Windows operator **I want** `winget install MailFathom.mfctl` to put `mfctl` on my `PATH`, **so that**
getting the administrative client is one command rather than a download, a checksum comparison, a rename, and a
`PATH` edit.

**As a** Windows operator who already has it **I want** `winget upgrade` to carry me to the next release, **so
that** I do not have to watch the releases page to learn that one happened.

**As** the owner **I want** the manifest submitted by the run that published the release, **so that** the version
winget offers is never one behind the version that exists.

## Acceptance

- `scripts/build-winget-manifests.sh` renders the three manifest files for a version into an output directory,
  reading each `InstallerSha256` from the bytes the release attaches rather than fetching the asset a second
  time. It touches no network and no credential, so the manifest set can be produced and read locally before
  anything is submitted.
- The rendered set validates against the `1.12.0` schema: every file carries its
  `# yaml-language-server: $schema=` header, the installer file declares `InstallerType: portable`,
  `Commands: [mfctl]`, and one entry per released Windows architecture, and the defaultLocale file names the
  publisher, the license, the repository, and the documentation site.
- A reusable workflow submits the set as a pull request to `microsoft/winget-pkgs` through
  `Krzysztof318/winget-pkgs`, syncing that fork before it branches, and opens exactly one package version per
  pull request as the community repository requires.
- `Release` calls it after `Publish the GitHub release`, because winget validation downloads the installer from
  the URL the manifest names and that URL does not resolve until the release exists. `Nightly` never calls it: a
  nightly build is not a version winget can offer.
- The job is guarded the way the command build is, so its failure leaves a published release intact and shows as
  a red job beside it rather than withholding anything.
- The submission credential is a repository secret, and `docs/operations/release-procedure.md` states which scope
  it carries, what that scope can reach, and why no narrower token can do the job — so the breadth is recorded
  where the next reader of the pipeline finds it rather than inferred from a secret name.
- `docs/users/installation.md` and `docs/operations/admin-endpoint.md` give the winget command where they
  currently describe downloading a binary, and say plainly that no binary is signed and what the checksum
  verifies. `README.md` names it wherever it names getting the command.
- No new action owner and no new tool: the submission uses `gh` and `git`, both already on the runner, so
  `THIRD_PARTY_LICENSES.md` gains no row and the approved-owner set in `scripts/test-agent-workflow.sh` is
  unchanged.

## Out of scope

- Any Linux package channel. #495 carries `apt`, with its own hosting, signing, and key-custody decisions.
- Scoop, Chocolatey, and Homebrew. Each is a separate submission contract and none is decided by this one.
- Code signing. #492 removes the pipeline that would have produced a signature, and winget accepts an unsigned
  portable package; SmartScreen will still warn on first run, and the documentation says so rather than
  implying otherwise.
- A winget package for the MailFathom service itself. Linux is the only runtime for the service, and #319 is
  where Windows for it is decided.

## Dependencies

- #492 settles that no `mfctl` binary carries a signature, which is what the manifest and the documentation
  state.

Krzysztof318:
Offers a released `mfctl` through the Windows Package Manager, so a Windows operator gets the administrative command
with `winget install MailFathom.mfctl` instead of a download, a checksum comparison, a rename, and a `PATH` edit.

The binaries already had the shape winget takes — one self-contained file per architecture, at a version-specific URL
that never moves — so what this adds is the manifest and the submission, and nothing about how the command is built.

## What it does

`scripts/build-winget-manifests.sh` renders the three-file manifest set at schema `1.12.0` under
`manifests/m/MailFathom/mfctl/<version>/`. A bare `.exe` is `InstallerType: portable` with `Commands: [mfctl]`, which
is what makes the command reachable under its own name rather than under the versioned file name a release attaches
it as; `jqlang.jq` is the same shape against the same kind of asset. Each `InstallerSha256` is computed from the bytes
`Build the CLI binaries` produced rather than from a download of the published asset. The script reaches no network and
reads no credential, so the set can be produced and read anywhere.

`Submit the winget manifest` opens the pull request against `microsoft/winget-pkgs`, through a fork it creates if it is
absent and force-syncs before branching. `Release` calls it last, after `Publish the GitHub release`, and that ordering
is a requirement rather than a preference: winget's validation downloads the installer from the URL the manifest names,
and that URL does not resolve until the release exists. `Nightly` never calls it. It gates nothing, like the build it
consumes.

The submission is idempotent the way the registry publications are: a version the community repository already carries
is reported and skipped, and a re-run whose branch already has an open pull request reports that rather than opening a
second one — which matters more here, since winget-pkgs allows exactly one pull request per package version.

## The decision worth reviewing

**`WINGET_PKGS_TOKEN` is a classic personal access token with the `public_repo` scope, and that scope is
account-wide.** Nothing narrower can do the job: a fine-grained token creates the commit on the fork and then fails on
the pull request, because it cannot be granted a permission on a repository this account does not own. That is the same
breadth that kept a classic `project` token out of this repository's secrets for writing the roadmap board, so it is
recorded rather than passed over — in `docs/operations/release-procedure.md` where the pipeline is read, and in
`THIRD_PARTY_LICENSES.md` beside the channel it reaches. The alternative was a manual submission per release, and the
owner chose the pipeline.

Two things follow that are not code. The secret has to exist before the next release, or the job fails naming it. And
Microsoft's open-source Contributor License Agreement has to be signed once by the account the token belongs to; it is
the third live obligation in the register now.

## Verification

- `scripts/verify-full.sh` — green on the rebased branch.
- `scripts/test-agent-workflow.sh` — 110 passed, including two new contracts over the generator: that the installer
  manifest names `portable`, `mfctl`, both architectures, and a hash taken over the built bytes, and that a missing
  Windows binary fails while the release run is still on screen rather than in somebody else's pull request days later.
- The rendered set was validated against the official `1.12.0` JSON schemas for all three manifest types.
- `actionlint` with its bundled `shellcheck` reports nothing on either changed workflow beyond the `SC2016` notes the
  repository's `printf` style already produces everywhere.
- The sparse-clone, branch, and commit flow was rehearsed against a local fixture repository, including the first-
  submission case where the package directory does not exist upstream.

## Documentation

`docs/operations/release-procedure.md` gains the channel, the ordering, the idempotence, and the credential.
`docs/operations/admin-endpoint.md` gains the winget path under *Getting the command*, stating plainly that no version
has been accepted yet and that winget verifying a hash is not a publisher check. `docs/users/installation.md` and
`README.md` point at it. One cell in the release-procedure artifact table still read `while signing is switched off`
after #492 removed signing; it is corrected here rather than left contradicting the page around it.

## Out of scope

`apt` is #495. Scoop, Chocolatey, and Homebrew are each their own submission contract. Code signing is settled by #492
and this changes nothing about it.

Closes #494


scripts/build-winget-manifests.sh
```diff
@@ -0,0 +1,183 @@
...
+#
+# The binaries directory is what `Build the CLI binaries` produced: the release's own bytes, before anything publishes
+# them. Each `InstallerSha256` is computed from those, rather than by downloading the asset back from the release page,
+# so the hash a Windows operator's `winget` compares against is the hash of the file this pipeline built. A download
+# would agree in every case except the one worth catching.
+#
+# Nothing here reaches the network and nothing reads a credential, so the set can be produced and read on any machine
+# holding the binaries. Submitting it is a separate act — see .github/workflows/submit-winget-manifest.yml.
+
+readonly package_identifier='MailFathom.mfctl'
```
Krzysztof318, line 27: this is id but what will be the name visible on list?

Krzysztof318: The name on a listing is `PackageName` from the defaultLocale manifest, not this identifier — `winget search` and `winget list` print it in their `Name` column with the identifier beside it. It read `mfctl`, so a row would have named the command and said nothing about what it administers.

Fixed in 1d05d5d: `PackageName: MailFathom CLI`. The command still reaches an operator by the two routes winget has for it — `Moniker: mfctl`, which `winget install mfctl` resolves, and `Commands: mfctl` in the installer manifest, which puts it on the `PATH`. A search now reads:

```
Name            Id                Version  Source
MailFathom CLI  MailFathom.mfctl  0.4.0    winget
```

The same commit moves `Publisher` from `Krzysztof Kasprowicz` to `MailFathom`. winget's convention is `Publisher.Package` and the two are read together, so a submission whose `Publisher` disagrees with the identifier it is filed under is a question somebody else's reviewer asks days later. `Author` keeps the person, and `PublisherUrl` now points at the documentation site rather than at a personal profile. That is the shape every comparable published package uses: `GitHub.cli` is `GitHub, Inc.` / `GitHub CLI` / `gh`, and `Kubernetes.kubectl` is `The Kubernetes Authors` / `Kubernetes CLI` / `kubectl`.

`winget_manifest_names_the_product_and_the_command` in `scripts/test-agent-workflow.sh` now asserts all four fields, so the listing cannot quietly go back to naming the command.

PR #510, opened 2026-08-06
[#500] Name MailFathom's own telemetry in one place and subscribe it in another

### LINKED ISSUE #500: Give MailFathom's own telemetry one naming convention and one registration point
Sub-issue of #85, and the one that gates the rest of it.

MailFathom emits no telemetry of its own. A search over `src/` for `ActivitySource`, `new Meter(`, `IMeterFactory`, `Counter<`, or `Histogram<` returns nothing, so every span and every instrument the process publishes today comes from a library: ASP.NET Core, `HttpClient`, the runtime, the Aspire Npgsql EF Core enrichment, and the `Polly` meter. `src/Host/ServiceDefaultsExtensions.cs` reserves a place for the first-party half with `tracing.AddSource(builder.Environment.ApplicationName)`, but nothing publishes to that source, so the line subscribes to an empty stream.

That is the shape this issue corrects, and it corrects nothing else: it adds the convention and the single registration point, not the spans and the counters. Every sibling issue then has one answer to *what is this source called* and *where is it subscribed*, instead of each feature inventing both. Doing it the other way round is how a repository ends up with `MailFathom.Sync`, `MailFathom.Mail.Synchronization`, and `Synchronization` naming the same thing.

## User stories

**As an operator**, I want every MailFathom-emitted span and instrument to carry a name I can predict from the part of the system it came from, so that a collector's filter or a dashboard query written against one boundary keeps working when the next boundary is instrumented.

**As a contributor**, I want one place that declares which of MailFathom's own sources and meters exist and one place that subscribes them, so that adding an instrument is a change in the boundary that owns it rather than an edit to the composition root that has to be remembered.

**As a privacy owner**, I want the tag-cardinality and content rules stated once beside the convention rather than restated per feature, so that a reviewer checks a new instrument against a written rule instead of against precedent.

## Acceptance

- A documented naming convention for MailFathom's own `ActivitySource` and `Meter` names, stated in one place, with the rule that decides the name from the boundary rather than from the feature.
- The names are declared where a reader finds them together rather than as string literals spread across the boundaries that emit them, and each boundary reaches its own source or meter through that declaration.
- One registration point subscribes the declared set, replacing the `AddSource(builder.Environment.ApplicationName)` placeholder, so a source that exists but is unsubscribed cannot merge unnoticed.
- Meters are obtained through `IMeterFactory` where the boundary is a DI-registered service, so a test can observe the instruments without process-wide state.
- The tag rules are written beside the convention and are the ones every sibling issue is reviewed against: no per-email, per-UID, per-address, per-subject, or per-content tag; account and folder appear as MailFathom's own configured aliases or not at all.
- Unit tests assert the declared names and that the registered set and the declared set are the same, so a source added without a subscription fails the suite.
- `docs/operations/telemetry.md` states the convention and what a reader can therefore expect a MailFathom instrument to be called.
- The unit-test suite and the 85% coverage gate pass.

## Depends on

- Nothing. This is the gate #502, #503, #504, and #505 wait on.

Krzysztof318:
Closes #500

Every span and instrument this process publishes today comes from a library: ASP.NET Core, `HttpClient`, the runtime, the Aspire Npgsql EF Core enrichment, and the `Polly` meter. The one line reserved for MailFathom's own half subscribed `builder.Environment.ApplicationName`, which nothing publishes to, so a subsystem about to emit its first span had neither a name to use nor a place that collects it — and inventing one per feature is how a repository ends up with three names for one boundary.

## What this adds

`Telemetry` in `Common` declares one name — `MailFathom` — and carries the one `ActivitySource` and the one `Meter` that bear it. A subsystem does not choose what it is called and does not construct a source or a meter of its own: it starts its spans on `Telemetry.ActivitySource` and creates its instruments on `Telemetry.Meter`. That is what makes a name invented for a feature impossible rather than merely discouraged — there is no second source to give a different name to.

There is no per-subsystem name, and none is added until there is something a name is the right way to tell apart. Which subsystem a signal came from is already carried by the span or instrument name and by its tags, where a distinction costs an operator nothing, while a second registration is one more thing to subscribe to before anything is collected.

One string serves both registries. An activity source and a meter are separate subscriptions to OpenTelemetry and cannot collide, so spans and instruments go under one name rather than two that could drift apart.

Neither instance is disposed. Both live for the lifetime of the process, disposing a shared source would silence every other publisher, and nothing is reclaimed by disposing either at shutdown — so a type that reports through them implements no disposal on their account.

`TelemetrySubscriptionExtensions` in `Host` subscribes that declaration rather than repeating the string, so the host and the publishers cannot disagree about it. That is the failure this is shaped against: an unsubscribed source produces no span however much code publishes to it, and nothing reports it.

The privacy and cardinality contract is written beside the convention rather than restated per feature — counts, sizes, durations, outcomes, error codes, and MailFathom's own configured aliases are permitted; mail content, an address, a subject, a remote folder path, a message identifier, a UID, a search term, a credential, and prompt or completion text are not.

## One acceptance item from #500 is deliberately not met

#500 asks that "meters are obtained through `IMeterFactory` where the boundary is a DI-registered service, so a test can observe the instruments without process-wide state." **This change does not do that, and no follow-up will.** The design settled in review is the opposite one: a single process-wide `ActivitySource` and `Meter` on `Telemetry`, which is what removes a publisher's ability to name itself. An `IMeterFactory` meter is named where it is resolved, so the factory route hands the naming decision back to each boundary — the exact thing #500 exists to prevent.

What that costs is real and worth stating: a unit test cannot observe these instruments through a provider of its own, and reads them through a process-wide `MeterListener` instead. The item is recorded as rejected on #500 rather than left to look unfinished behind a closed issue.

## The one name that already existed

**Breaking, against the telemetry surface.** #490 merged while this was open and instrumented mailbox mutations under a name of its own, `MailFathom.Mailbox.Mutations`, subscribed directly from `ServiceDefaultsExtensions`. It now publishes through the shared source and meter like everything else.

Nothing about what is emitted changes: the same spans, the same counter and histogram, the same mutation, account, folder-alias, and outcome tags. Only the name they arrive under does, and which mutation ran was already a tag rather than part of the name.

**Operator action:** a dashboard, saved query, or alert rule filtering on `MailFathom.Mailbox.Mutations` selects nothing after this and has to name `MailFathom`. No configuration key, tool argument, database column, or deployment value moves, so nothing outside an observability backend needs touching.

## A suppression this forced

`MailboxMutationTelemetry` stopped being `IDisposable`, and that woke CA2000 on `MailKitImapWriteSessionFactory.OpenForWritingAsync` — a file this change does not otherwise touch. The causality was verified by re-adding `IDisposable` and rebuilding, not assumed.

No arrangement of that method satisfies the rule: `MailboxWriteConnectionLease` releases asynchronously and implements no synchronous `IDisposable`, which is the call CA2000 asks for by name. The expression form and an exception-safe `try`/`catch` were both tried and both still fired, and the `catch` was unreachable anyway, since the session constructor only assigns fields. It is therefore a `[SuppressMessage]` on that one method, with the reasoning written into the justification.

## Tests

- `TelemetryTests` pins the published string, because the name reaches a dashboard filter and an alert rule, and asserts that both the activity source and the meter carry it — an instance under some other name is the same silent failure arriving from the other side.
- `TelemetrySubscriptionExtensionsTests` asserts the subscribed name against the declared one through recording builders, hand-written because both OpenTelemetry builders are abstract classes.

## Notes

- `Host` gains an explicit `ProjectReference` to `Common`, so `docs/architecture/solution-structure.md` gains the arrow and the third consumer; the diagram there states that it is the whole set of references.
- The only `packages.lock.json` change is the `MailFathom.Common` line under `MailFathom.Host` — a first-party project, not a dependency pin, so `THIRD_PARTY_LICENSES.md` is unaffected.
- No test names `ServiceDefaultsExtensions`, which `scripts/review-obligations.sh` reports. The decision that moved there is which subscriptions are composed, and that is what the Host test asserts; `tests/AGENTS.md` keeps `Host.UnitTests` away from asserting dependency-injection wiring.
- ADR 0007's `describes:` marker covers `src/Infrastructure/Observability/**`, so it is reported too. It records the write-session boundary and the three channels a mutation reports on, and names neither the meter nor the activity source, so the change above leaves it accurate.


src/Common/Observability/MailFathomTelemetry.cs
```diff
@@ -0,0 +1,53 @@
...
+/// <para>
+/// A meter is obtained from <see cref="System.Diagnostics.Metrics.IMeterFactory" /> by the service that owns the
+/// instruments, so a test observes them through its own provider instead of through process-wide state. An activity
+/// source has no such factory and is created directly from the name below.
+/// </para>
+/// </remarks>
+public static class MailFathomTelemetry
+{
+    /// <summary>The prefix every name below carries.</summary>
+    public const string NamePrefix = "MailFathom";
```
Krzysztof318: I would stay with MailFathom namespace only for now.

Krzysztof318: Done in 4ad71d4 — there is one name now, `MailFathomTelemetry.Name = "MailFathom"`, and `NamePrefix` plus the four subsystem constants and `All` are gone with it. `MailboxMutationTelemetry` publishes under it, the host subscribes it, and the docs page carries one name instead of a table.

Recording the trade-off rather than disputing it: an operator can no longer select one subsystem by source name, so telling mailbox work from MCP work is a filter on the span or instrument name and its tags. That reads fine while one subsystem is instrumented, and splitting later is a name change under the same prefix, which 0.x permits.

Note that this makes the fold I did during the rebase moot in passing: #490 had merged meanwhile with a name of its own, `MailFathom.Mailbox.Mutations`, subscribed directly from `ServiceDefaultsExtensions`. It now publishes under `MailFathom` like everything else, and `MailboxMutationTelemetryTests` fails if a publisher invents a name again.

docs/operations/telemetry.md
```diff
@@ -21,19 +21,49 @@ its attempts, outcomes, timeouts, and circuit-breaker state transitions.
...
+| `MailFathom.Persistence` | Local storage: the email content store and the write sessions around it |
+| `MailFathom.Extraction` | Mail text extraction and the backfill that reprocesses what earlier runs left |
+
+The name is decided by the subsystem, never by the feature that happens to emit first and never by the assembly the
+code sits in — that is what an operator filters a dashboard on, and what survives a type moving between projects. The
+shared `MailFathom.` prefix is what lets one filter select everything this process owns and nothing a library emits.
+All four are subscribed by the host together, from one declaration, so a name that exists without being collected is a
+failing test rather than a silently empty stream. Instruments that are collected in aggregate come from the meter
+factory the owning service is given, rather than from a process-wide instance, which is what keeps them observable in a
+test.
```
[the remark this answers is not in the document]
Krzysztof318: Confirmed and fixed in 4ad71d4. `grep -rn IMeterFactory src/` returned only that doc comment, and `MailboxMutationTelemetry` builds `new Meter(MeterName)` in its own constructor and disposes it alongside its `ActivitySource`, so the sentence described a convention with no adherents.

Removed rather than scoped. A forward-looking rule whose only candidate does not follow it is the same defect one release later, and how a publisher obtains its meter follows from the lifetime it is registered with — which is the publisher's decision to record, not this type's. The page and the `MailFathomTelemetry` remarks now say only what the declaration is: the name.

The surrounding text moved anyway — the owner asked for a single `MailFathom` name, so the four-name table and the "three of the four are quiet" framing are gone too.

src/Common/Observability/MailFathomTelemetry.cs
```diff
@@ -0,0 +1,39 @@
...
+/// which is a cardinality rule as much as a privacy one, because every one of those would open a time series per
+/// message or per person.
+/// </para>
+/// <para>
+/// What this declares is the name and nothing else. How a publisher obtains its meter and its activity source, and
+/// which of them it disposes, stays with the publisher, because that follows from the lifetime it is registered with
+/// rather than from what it is called.
+/// </para>
+/// </remarks>
+public static class MailFathomTelemetry
```
Krzysztof318: short to Telemetry also add public static Activity source and metrics source here

Krzysztof318: Done in 56935ce. `MailFathomTelemetry` is now `Telemetry` (file renamed with it), and it carries `public static ActivitySource ActivitySource` and `public static Meter Meter` beside `Name`.

`MailboxMutationTelemetry` creates neither now — it starts its spans on `Telemetry.ActivitySource` and builds its counter and histogram on `Telemetry.Meter`. That turns the convention into something structural: there is no second source to give a different name to, so a name invented for a feature is impossible rather than discouraged.

Two consequences worth naming. `MailboxMutationTelemetry` stopped being `IDisposable` — disposing a shared source would silence every other publisher — so its test harness and the integration test lost that disposal too. And CA2000 woke up on `MailKitImapWriteSessionFactory.OpenForWritingAsync` once that type stopped being disposable; I verified the causality by re-adding `IDisposable` and rebuilding. No arrangement of that method satisfies the rule, because the lease releases asynchronously and implements no synchronous `IDisposable`, which is the call CA2000 asks for by name. I tried the expression form and an exception-safe `try`/`catch` first; both still fired, and the `catch` was unreachable anyway since the session constructor only assigns fields. It is a narrow `[SuppressMessage]` on that one method with the reasoning written out.

Krzysztof318:
Answering the P2 from the last review, since a finding in a review body has no thread to resolve.

The finding is right that #510 does not meet #500's `IMeterFactory` acceptance item, and right that closing the issue would bury it. It is not deferred, though — it is rejected, and the reason is the design that landed since: `Telemetry` now holds one process-wide `ActivitySource` and one `Meter`, and every publisher reports through those. A meter from `IMeterFactory` is named where it is resolved, so that route hands the naming decision back to each boundary, which is what #500 exists to prevent.

Recorded in the pull-request body under *One acceptance item from #500 is deliberately not met*, and on the issue at https://github.com/Krzysztof318/MailFathom/issues/500#issuecomment-5216326706, including what it costs: a unit test reads these instruments through a process-wide `MeterListener` rather than through a provider of its own.

Issue #452, opened 2026-08-05
[P] Let MailFathom write to the remote mailbox


Krzysztof318:
## What this feature is

MailFathom gains the ability to change a mailbox it synchronizes: to relocate a message between folders, to delete one on the server, to mark one read or unread, and to copy one into a second folder. It is done when a change MailFathom performs is indistinguishable, from the mailbox's side, from the same change made by hand in a mail client — and distinguishable from it in exactly one place: MailFathom's own records, which know it was ours.

Writing is a capability rather than a use of one, which is why it stands beside the rule engine rather than inside it. #251 is its first consumer, but a future MCP tool and the mail-client gateway sketched in draft section 21.7 are consumers too, and each would otherwise have to solve occurrence identity and loop prevention again.

The read guarantee is not being traded away for it. What changes is the invariant's shape: a read path — synchronization, content retrieval, any MCP tool — stays structurally incapable of writing, while an act the owner authored may write. That reading is what lets root `AGENTS.md` stand verbatim through the whole feature, since its sentence is scoped to retrieval; draft section 11.1 states the invariant unscoped and is amended with #447.

`\Seen` is the mutation that shows what the distinction is worth. Marking mail read is now permitted as an authored act and remains impossible as a side effect of reading, and the stored flag keeps exactly one writer — synchronization observing the server — because an authored change is a request recorded rather than a local edit.

Almost all of the remaining difficulty is in one mechanism. A move without the MOVE extension is three commands and a crash can land between any two of them; a `COPY` issued twice is a second message rather than a repeat of the first; the message in the target folder is a different occurrence with a different UID; and every one of these changes comes back through synchronization looking like something to react to. One durable mutation record, written before the first command, answers all of it — idempotency, identity, and provenance — which is why it gates most of what follows.

## What is refused

#446 settled the boundary and is closed. Sending is outside this feature and stays there: send, reply, and forward belong to the SMTP outbox and its own authorization review. So does every flag other than `\Seen`, and so does folder management. Permitting one of them later reopens that decision rather than reading a gap as permission.

## Order

| # | Role |
|---|---|
| #447 | Gate. The session every other child acts through, the type separation that keeps the read path unable to write, and the ADR and amended draft section 11.1 that record the end of the never-writes posture. |
| #448 | The mutation record. Every consumer and every action below is a reader of it, so it lands before all of them. |
| #449, #450 | Either order once #448 exists. Identity and provenance are independent readers of the same record. |
[...]

Issue #445, opened 2026-08-05
[P] ask_mail: cited answers with the model as a second filter


Krzysztof318:
## What this feature is

MailFathom answers questions about a mailbox instead of only finding things in it. `ask_mail` is the fourth MCP tool from draft section 13.4: a question in, an answer out, every claim cited to a stable local message identifier the caller can then fetch. The retrieval underneath it is the hybrid search built by #436; what this feature adds is the agent that uses it, the framing that makes mail safe to hand a model, the second filter that decides which retrieved passages actually answer the question, and the bounds and evidence that make running any of it accountable.

The second filter is the part worth naming separately, because it is where the model earns its cost. Fused lexical and vector ranking knows which passages resemble a question; it does not know which of them answer it. Putting a bounded candidate set to the model with a fixed output schema, and dropping what falls below a threshold, is what separates ten messages that mention a claim from the one that states what was agreed. It is optional, and disabled it costs nothing and changes nothing.

Everything here is gated on providers being configured and healthy. Draft section 12.2 exposes the tool only under that condition and hides it otherwise, so an instance with no chat provider is not an instance with a broken tool — it is an instance with three tools, exactly as today.

Two properties hold across the whole feature and neither is a limitation to be relaxed later. Mail reaching a model is untrusted input, structurally separated from instructions and never treated as a command. And the tool has no mutation capability to expose: it cannot send, delete, move, or mark mail as read, and that stays true however much the rest of MailFathom later gains the ability to write.

## Order

| # | Role |
|---|---|
| #437 | Gate. Nothing above it exists without a chat client, and its health is what decides whether the tool is advertised at all. |
| #438 | The composition. Needs #437 and the hybrid retrieval from #436. |
| #439 | Gate for everything that reads mail through a model. #440 and #441 both rest on it, so it lands before either. |
| #440, #441 | Either order once #439 exists. The filter and the tool are independent; the tool works without the filter, and the filter improves any retrieval it is given. |
| #442 | Needs both of the above, because a ceiling is only meaningful once there are two things spending against it. |
| #443 | Needs #441 — the run it records. |
| #444 | Closing. It checks the properties #439 and #440 claim, and it is deliberately not written by the change that claims them. |

## Done when

- An instance with healthy chat and embedding providers advertises `ask_mail`, and one without them advertises exactly the three tools it does today.
- Every answer cites message identifiers that resolve, and no answer can be produced from mail outside the caller's scope.
[...]

Issue #436, opened 2026-08-05
[P] Semantic retrieval over synchronized mail


Krzysztof318:
## What this feature is

MailFathom gains semantic retrieval over the mail it already stores. Today `search_emails` matches words: a message about "water damage" is not found by a search for "the roof leak", because PostgreSQL full-text search has no way to know they are the same subject. This feature closes that gap end to end — text is chunked, chunks are embedded through a configured provider, vectors are stored and indexed per profile, and the existing search tool fuses vector similarity with the lexical rank it already computes.

It is done when an instance with a provider configured answers a question by meaning as well as by word, and an instance without one behaves exactly as it does today. That second half is not a fallback; draft section 12.2 makes provider-neutral operation a property of the product, and this feature is where it first has to be true rather than merely stated.

The whole of it is invisible in the tool contract. `search_emails` keeps its arguments, its result shape, its annotations, and its bounds; what changes is which messages come back and in what order. A client written against the lexical tool keeps working when its instance grows a provider, which is what makes enabling this an operator's decision rather than a migration for everyone who calls it.

`ask_mail` is not here. It consumes this feature and is its own parent, because a cited answer over retrieved mail raises questions — egress, cost, prompt injection — that retrieval itself does not.

## Order

| # | Role |
|---|---|
| #424 | Gate, answered by ADR 0006. The profile is the geometry of a vector space and is immutable; the chunk hash carries the boundary rules; the profile row is the generation; configuration declares the model, provider, and fallback chain while activation is what spends against them; and activation states its estimate first. The schema, the adapter, the chunk hash, the reindex, and the commands all encode that record. |
| #425, #426, #427 | Any order once #424 lands. Chunks, tables, and the provider adapter are independent of each other. |
| #428 | The first thing that joins them, and the point at which new mail starts being embedded. |
| #429, #430, #434 | Any order once #428 exists: existing mail, the index, and the ceilings on what all of it may spend. |
| #431 | What makes the feature visible to a caller. Needs #428; better with #430. |
| #432 | Pairs with #431 — the degraded path it describes only exists once retrieval has a semantic half to lose. |
| #433 | Needs #429 and #430, and is what turns a model change from an outage into an operation. |
| #435 | Closing. It is the operator's way into everything above, so it lands once there is something to activate, reindex, and report on. |

## Done when

- An instance with an active healthy profile answers `search_emails` by meaning and by word, with unchanged arguments and result shape.
[...]

Issue #405, opened 2026-08-04
[P] Build the durable job model that mail automation and classification both consume


Krzysztof318:
## What this feature is

MailFathom gains durable background work: a job that is enqueued against committed local state, claimed by one worker, retried only where retrying is safe, bounded in how much of the instance it may consume, and visible afterwards. It is done when a process can be killed mid-execution and the work it held is neither lost nor repeated, and when a job that can never succeed is a thing the owner can see and act on rather than a load they have to notice.

Draft section 21.5 describes this subsystem and nothing has built it. [#251](https://github.com/Krzysztof318/MailFathom/issues/251) and [#76](https://github.com/Krzysztof318/MailFathom/issues/76) each named that absence as their first blocker when this was written, so whichever of them started would have built it on the way to something else — and the second would then have had a choice between adopting a mechanism shaped around the first feature's needs or bringing its own. A second scheduler beside the first is the failure mode this feature exists to prevent, which is why the mechanism is a feature with one owner rather than a side effect, and why it stands before its consumers rather than inside one of them.

**#251 then answered that question differently, and the reasoning is worth carrying here rather than leaving this issue to describe a blockage that no longer exists.** #455 put rule evaluation in the account synchronization run — which already has per-account isolation, jittered backoff, a slot count that stops one account starving another, and a failure path that defers the account rather than the process — and refused this feature as a dependency. #453–#458 shipped that way. What decided it is a criterion this feature is best read through: **work belongs here when it needs its own backoff per item, and in the loop that already recurs when it does not.** NCalc evaluation over committed local state has no transient failure, so it carries no attempt state and needs no job row. A call to a sidecar has both. So this is no longer a mechanism shared by two features, and the argument for building it on its own is now the narrower one: without it, a consumer grows an attempt count and a next-attempt time inside its own record, with no lease, no terminal state, and nothing to observe.

What is in scope is the mechanism and not a use of it. Rule definitions, classification, and anything that reaches the remote mailbox belong to the issues that own them; this feature has no consumer of its own and ships with none.

Two boundaries come from the existing architecture rather than from this feature, and both hold across every child. A job is enqueued after a message and its required content are committed — never inside the IMAP synchronization transaction — so a provider retry or a worker crash cannot produce a different processing boundary. And no job runs on a path an MCP read waits on, because an MCP request must not become slower or fail because automation is behind. Neither can be proved here, since nothing enqueues yet; what this feature does is make them structural, and the first consumer is where they are demonstrated.

[...]

PR #512, opened 2026-08-06
[#508] Give a parent its own Queue value, a title prefix, and a stated nesting depth

### LINKED ISSUE #508: Give a parent issue its own Queue value, a title prefix, and a stated nesting depth
A parent issue is not a unit of work, but the board has no value that says so. `Queue` is set on every open issue, and each of the four existing values reads wrongly on a parent whose children are spread across releases: `Later` claims accepted scope not yet started while its children are in flight, `Needs decision` claims it waits on an answer, `Parked` claims no commitment, and `Next` puts a container into the one view that is meant to hold what the owner is working on now — spending one of the five slots that decision uses.

Two smaller things belong with it, because all three are how a parent is recognised. Nothing in a title says an issue is a parent, so a reader gets that only from the `parent` label or from opening it. And the hierarchy has no stated depth: the rules today describe one feature split into several issues and say nothing about whether a child may have children of its own, which leaves the answer to whoever links the next one.

Four further changes joined this issue while it was being worked, because all of them are the same page's contract and splitting them would have left it describing a board that no longer existed. `Status` gains `Ready to merge`, since the two built-in review workflows are being enabled — approval moves an item into it and a review requesting changes moves it back — and the board's own README had drifted far enough to be misleading: it still named the `Track` field that became `Area`, listed its deleted `A`–`E` values, and knew only five of the eight views. `Size` stops being deferred to a planning pass that never happens separately here and is estimated when the issue is opened. And the role rules stop inferring board access from the remote: the owner grants read or write on project `4` to a contributor whenever they decide to, so a fork is no evidence of no access, and the skills probe for it instead.

## User stories

- As the owner reading `Now`, I want the view to hold only work I chose to start and work a pull request carried in, so a cross-milestone parent never occupies a slot it cannot be worked in.
- As the owner scanning any board view or issue list, I want a parent to read as one from its title alone, without opening it or checking its labels.
- As the owner reading `Now`, I want an approved pull request to sit apart from one still being written, so the column an item is in says whether I am being asked to review it or to merge it.
- As an agent linking sub-issues, I want the permitted depth of the hierarchy stated, so a feature large enough to need an intermediate grouping gets one and nothing deeper is invented.

## Acceptance

- `Queue` carries a fifth option, `Parent`, added without clearing any assignment held by the four that exist.
- A parent issue whose children span releases — the one that carries `cross-milestone` and no milestone — takes `Queue: Parent`. A parent that fits a single milestone keeps the ordinary values, so `Next` and `Later` mean on it exactly what they mean anywhere else.
- Every parent issue's title is prefixed `[P]`, and the existing parents are brought onto it.
- Sub-issues may nest exactly one level further: a parent may have a sub-parent, and a sub-parent's children are leaves. A sub-parent is a parent in every rule — the `parent` label, the `[P]` prefix, and the same `Queue` rule.
- `Status` carries `Ready to merge` between `In progress` and `Done`, and the page records both review workflows as the pair they are, so the value tracks the newest verdict rather than the first.
- The board's README and description describe the board that exists: `Area` and its nine values, all five `Queue` values, all eight views, the parent conventions, and the status transitions including the two review ones.
- `docs/operations/issue-tracking.md` states all of it, and the workflow files that write or describe `Queue` agree with it: `$start-task`'s placement step and `$finish-change`'s `Queue: Next` write.
- `Size` is set when an issue is opened, corrected against the measured diff at the pull request, and a parent carries `XL`. Every open issue that had no `Size` gets one.
- Board access is probed rather than inferred, in either role, through `projectV2(number: 4) { viewerCanUpdate }`, and the rules split it from repository access: a grant on the board never carries the label, the milestone, or the push. `$get-started-contributors`, `CONTRIBUTING.md`, `AGENTS.md`, and `docs/operations/agent-workflow.md` agree on that.
- The six open cross-milestone parents on the board are moved onto the new value, and no view filter has to change for `Now` to stop showing them.

Krzysztof318:
Closes #508

`Queue` gains a fifth value, `Parent`, for a parent issue whose children span releases. Each of the four that exist said something untrue about one — `Later` that nobody had started scope already in flight, `Needs decision` that an answer was missing, `Parked` that the project was not committed, and `Next` that the owner was working on a container, which also spent one of the five slots that decision uses. A parent that fits a single milestone is not this case and keeps the ordinary values. No view filter changed: `Now` asks for `queue:Next`, so a value that is not `Next` is already outside it.

Two smaller conventions join it, because all three are how a parent is recognised. A parent's title begins `[P]`, since an issue list, a search result, and a reference from another body all show a title on its own. And the hierarchy now has a stated depth — at most two levels, so a feature large enough that one of its parts split again gets a sub-parent and nothing invents a third. A sub-parent is a parent under every rule.

`Status` gains `Ready to merge` between `In progress` and `Done`, for the two built-in review workflows now enabled: approval moves an item into it and a later review requesting changes moves it back, so the value always reflects the newest verdict. `Now` then reads left to right as start it, finish it, merge it.

`Size` stops waiting for a planning pass that never happens separately here. It is estimated when the issue is opened, corrected against the diff the pull request measured, and a parent carries `XL` because its size is the sum of its children rather than a diff of its own.

Board access stops being inferred from the remote. The owner grants read or write on project `4` to a contributor whenever they decide to, so a fork is no evidence of no access and a clone of MailFathom made without write access is no evidence of having it. Both skills now probe `projectV2(number: 4) { viewerCanUpdate }` in either role, and repository access is separated from board access throughout: a grant on the board never carries the `type:*` label, the milestone, or the push. A negative probe is checked against the credential before it is believed, because a missing `project` scope, a `GH_TOKEN` in the environment, and a fine-grained token each fail identically to having no access at all — and that answer is what gets written into a `CLAUDE.local.md` and repeated in every later session.

## Breaking changes

None of the four public surfaces. This is the workflow contract and the private roadmap board.

## The board changes that came with it

The pull request cannot carry these, so they are already applied:

- `Queue` carries `Parent`, added with every existing option's id so all 254 assignments survived — verified by diffing all 292 items before and after, where the only differences were the six parents moved onto the new value;
- `Status` carries `Ready to merge`, added the same way;
- the six open cross-milestone parents (#85, #251, #405, #436, #445, #452) hold `Queue: Parent`, and every parent issue's title carries `[P]`, including the two closed ones so the convention has no historical exception;
- the 81 open items that had no `Size` have one, and every parent is `XL`;
- the board's README and description described a board that no longer existed — the `Track` field that became `Area`, its deleted `A`–`E` values, five views out of eight, and `Release 0.1.0`. Both are rewritten against this page.


Krzysztof318 (commented):
# NEEDS CHANGES

*A manual pass with the `Fathom review` reviewer prompt, run locally because GitHub Actions is unavailable today. Same inputs the workflow collects — the base checkout at `9548cac`, the changed files and their head content, `index-obligations.sh`, and issue #508 — plus two things the workflow's reviewer cannot reach: the live board and the project's built-in workflows.*

All seven files were read in full, along with the acceptance list of #508 and `obligations.json` (no tests, no registers; `docs/operations/agent-workflow.md` is the only page whose `describes:` marker covers a changed path, and the change touched it). There were no prior reviews, threads, or comments to reconcile.

The board confirms every claim under **The board changes that came with it**: `Queue` carries five options including `Parent`, `Status` carries `In progress` and `Ready to merge`, the six open cross-milestone parents sit on `Parent` and are all `XL`, all eight parent issues — the two closed ones included — carry `[P]`, and not one of the 92 open items is missing `Size`, `Queue`, or `Area`. The board's README and description match this page.

**Findings** — P2: 6, P3: 1

### Below the bar, for the record

- `Auto-add sub-issues to project` is enabled on the board and is absent from the transition list in `docs/operations/issue-tracking.md`. It sets no status beyond what `Item added to project` already does, but it is a workflow that starts meaning something now that the hierarchy has a second level.
- `docs/operations/issue-tracking.md` line 116 has no blank line before `The built-in workflows set ...`, so Markdown reads it as a lazy continuation of the paragraph above — previously the `Size` bullet, now the new paragraph about a parent's `XL`. The missing blank line predates this change; the page is published by docfx.
- The rewrapping in `.agents/skills/start-task/SKILL.md` lines 100–106, `.agents/skills/finish-change/SKILL.md` line 55, and `docs/operations/agent-workflow.md` lines 200–202 leaves those paragraphs ragged against the width each file otherwise keeps. Nothing in the repository states a width and no linter checks one, so this is an observation rather than a finding.
- Board hygiene, unrelated to this change: 15 open items hold `Queue: Next`, six of them in `Todo`, against the five-slot cap the page states for the owner's own writes.

docs/operations/issue-tracking.md
```diff
@@ -182,8 +190,9 @@ Writing it from the skill is not a shortcut past the automation; it is the only
 
 ## Status transitions
 
-- The board's `Status` field has `Todo`, `In Progress`, and `Done`.
-- The board's built-in workflows own every transition: `Auto-add to project` places a newly opened issue on the board, `Item added to project` puts it in `Todo`, `Pull request linked to issue` moves it to `In Progress`, and `Pull request merged`, `Auto-close issue`, and `Item closed` carry it to `Done`. Do not set those statuses by hand; a manual status that contradicts the automation hides the real state.
+- The board's `Status` field has `Todo`, `In progress`, `Ready to merge`, and `Done`.
+- The board's built-in workflows own every transition: `Auto-add to project` places a newly opened issue on the board, `Item added to project` puts it in `Todo`, `Pull request linked to issue` moves it to `In progress`, `Pull request approved` moves it to `Ready to merge`, `Pull request has a review requesting changes` moves it back to `In progress`, and `Pull request merged`, `Auto-close issue`, and `Item closed` carry it to `Done`. Do not set those statuses by hand; a manual status that contradicts the automation hides the real state.
```
Krzysztof318: **P2 — Name the two review workflows as the board names them**

**Impact**

This sentence lists the built-in workflows verbatim, and every other name in it is exact — `Auto-add to project`, `Item added to project`, `Pull request linked to issue`, `Pull request merged`, `Auto-close issue`, and `Item closed` all match. The two new ones do not exist. Querying `projectV2(number: 4) { workflows }` on project `4` returns `Code changes requested` and `Code review approved`. A reader checking whether the automation is enabled will look for `Pull request approved` in the board's automation list and not find it.

**Correction**

`Pull request approved` becomes `Code review approved`, and `Pull request has a review requesting changes` becomes `Code changes requested`. The board's own README repeats both wrong names under *Status is not yours to set* and is covered by the same acceptance item, so it takes the same correction.

**Rule** — `AGENTS.md`, "Documentation and test obligations": stale guidance is a defect, and this page is the only place these transitions are written down.

.agents/skills/start-task/SKILL.md
```diff
@@ -76,29 +95,35 @@ error rather than a partial result.
...
+9. Place the issue. The label and the milestone need write access to the repository, so they belong to
+   the owner's checkout; the board fields need write access to the board, which step 2 established.
+   It carries exactly one `type:*` label, an `Area` and a
    `Queue` value on the board, a milestone when the milestone rule assigns one, and a `Size` value
-   once the work is planned. Decide each from the rules on that page rather than asking. `Queue: Next`
+   estimated from the scope the body describes. Decide each from the rules on that page rather than asking. `Queue: Next`
    is never one of them: the owner chooses it, and `$finish-change` writes it once the pull request
-   exists, so a new issue takes `Later`, `Needs decision`, or `Parked`. Verify the values landed,
+   exists, so a new issue takes `Later`, `Needs decision`, or `Parked` — and `Parent` where the issue
+   being opened is a parent whose children span releases, which is also where its title takes the
```
Krzysztof318: **P2 — Separate the `[P]` prefix from the `Queue: Parent` condition**

**Impact**

This step reads `Parent` where the issue being opened is a parent whose children span releases, *which is also where its title takes the `[P]` prefix*, so the prefix is tied to the cross-milestone case. `docs/operations/issue-tracking.md` states it unconditionally — "A parent's title begins `[P] `" — and so does the acceptance list of #508. The board already works the unconditional way: #332 and #140 are parents that fit a single milestone, they hold ordinary `Queue` values, and both carry the prefix. An agent opening such a parent follows this step rather than the page and omits the prefix in what is the ordinary shape of a parent.

**Correction**

Split the two clauses: `Parent` where the children span releases, and the `[P]` prefix on every parent whatever its `Queue` value.

**Rule** — `docs/operations/issue-tracking.md`, "The issue that governs a change", and the acceptance list of #508.

docs/operations/issue-tracking.md
```diff
@@ -105,8 +107,12 @@ The board carries three single-select fields beyond `Status`. Set `Area` on ever
...
+
+  A parent issue takes `XL` and keeps it. Its size is the sum of its children rather than a diff of its own, and that sum is what puts it over the threshold, so the value reads as *this is delivered in pieces* on exactly the issues that already are. The warning `XL` carries elsewhere — split this before starting — is answered on a parent by the children themselves, which is what makes it the one place the value is not a problem to solve.
 The built-in workflows set `Status` and nothing else, so a newly opened issue reaches the board with no `Area` and no `Queue`. Setting both is part of opening the issue:
```
Krzysztof318: **P2 — Carry `Size` into this paragraph and into the audit query**

**Impact**

Line 91 now requires `Size` when the issue is opened, but this paragraph still says a new issue reaches the board with no `Area` and no `Queue` and that setting *both* is part of opening it, the audit below says to audit *both*, and its query filters `select(.area == null or .queue == null)`. An agent that follows the page literally sets two fields, passes the audit, and never sees a missing `Size` — which is exactly the omission this change had to repair by hand across 81 items.

**Correction**

All three fields in both sentences, and `select(.area == null or .queue == null or .size == null)` in the audit query.

**Rule** — `AGENTS.md`, "Documentation and test obligations": stale guidance is a defect.

CONTRIBUTING.md
```diff
@@ -69,14 +69,24 @@ The roadmap board, project `4`, is private to the maintainer and I have no acces
...
+
+An error naming permission from a credential that does carry the scope is the paragraph as written above. `viewerCanUpdate: false` makes it *the maintainer has granted me read access to project `4`: read it for context, and report every write as unavailable rather than attempting it*, and `true` makes it *the maintainer has granted me write access to project `4`, so the board steps in `start-task` and `finish-change` apply to me as written*. Nothing else in the block changes: your branch is still yours, your push still goes to your fork, and access to a board is not authority over a repository.
```
Krzysztof318: **P2 — The refusal is `NOT_FOUND`, not an error naming permission**

**Impact**

Run against a project the caller cannot reach, the probe answers `{"data":{"user":{"projectV2":null}},"errors":[{"type":"NOT_FOUND", ... "Could not resolve to a ProjectV2 with the number 1."}]}`, and `gh` prints one line: `gh: Could not resolve to a ProjectV2 with the number 4.` An error naming permission is what a *missing `project` scope* produces, which these texts already treat separately as a credential to repair. So the enumeration here has three outcomes and the answer a contributor without access actually gets matches none of them; the line above then sends them off to check credentials that are fine. `start-task` and `get-started-contributors` are saved by their "or a `null` project" clause, but they name the error the same wrong way, and `gh` shows its own message rather than the response body by default. `docs/operations/issue-tracking.md` line 9 and `docs/operations/agent-workflow.md` line 184 carry the same phrase.

**Correction**

In each of the five places, `an error naming permission` becomes a `NOT_FOUND` error — `Could not resolve to a ProjectV2 with the number 4` — with `projectV2: null`, and this file gains that answer as a row of its own.

**Rule** — `AGENTS.md`, "Documentation and test obligations", and the acceptance item of #508 requiring these four files to agree.

AGENTS.md
```diff
@@ -44,9 +44,9 @@ This file is loaded into every agent session, so it holds what has to be true be
...
 
-**The owner's checkout** is where `origin` is `Krzysztof318/MailFathom` and project `4` answers. Everything here applies, including the rules that exist nowhere else: the `agent/<short-description>` branch name, the linked worktree, the `type:*` label and the milestone on an issue, the board's `Area`, `Queue`, and `Size` fields, and pushing to this repository.
+**The owner's checkout** is where `origin` is `Krzysztof318/MailFathom` and project `4` answers. Everything here applies, including the rules that exist nowhere else: the `agent/<short-description>` branch name, the linked worktree, the `type:*` label and the milestone on an issue, and pushing to this repository. The board's `Area`, `Queue`, and `Size` fields are always written here and sometimes elsewhere, for the reason the next paragraph gives.
```
Krzysztof318: **P2 — Stop defining the owner's checkout by board access**

**Impact**

The owner's checkout is still "where `origin` is `Krzysztof318/MailFathom` **and project `4` answers**", which is the inference this change removes everywhere else. `$start-task` step 1 resolves the role from the remote alone and step 2 probes the board as a separate fact, so a clone of MailFathom without board access is the owner's checkout by the skill and is not by this sentence. This file is loaded into every session, which gives it the widest reach of the seven, and it contradicts the sentence added two lines below it.

**Correction**

Drop `and project `4` answers`; the rest of the line already defers the board to the next paragraph.

**Rule** — `AGENTS.md`, "The two roles this contract is written for", against `.agents/skills/start-task/SKILL.md` steps 1 and 2.

.agents/skills/finish-change/SKILL.md
```diff
@@ -52,29 +52,33 @@ Do not proceed while a gate fails.
    through `gh api repos/<owner>/<repo>/pulls/<number> -X PATCH -f body=...`. That endpoint names the
    *base* repository even for a fork's pull request, and it is the author who is allowed to patch it
    rather than anyone with write access to that repository.
-7. **Owner's checkout only:** set `Queue: Next` on the issue the pull request closes, whether it was
+7. **Wherever the board probe in `$start-task` returned write access:** set `Queue: Next` on the issue the pull request closes, whether it was
```
Krzysztof318: **P2 — Correct the opening sentence about what the role splits**

**Impact**

The file opens with "the steps below split on it in two places: the branch name, and the board". After this change the board follows the probe rather than the role — as this very step now says — while the role still decides the push in Publish step 4 and the target repository in step 5, neither of which that sentence names. A reader gets the inverse of the file's own structure.

**Correction**

"…split on it in two places: the branch name and the remote pushed to; the board follows the probe `$start-task` ran."

**Rule** — `AGENTS.md`, "Documentation and test obligations": stale guidance is a defect.

docs/operations/issue-tracking.md
```diff
@@ -105,8 +107,12 @@ The board carries three single-select fields beyond `Status`. Set `Area` on ever
...
... while its children are in flight, `Needs decision` that an answer is missing, `Parked` that the project is not committed to it, and `Next` that the owner is working on it, which also spends one of the five slots below on a container. A parent whose children all ship in one release is not this case and takes the ordinary values, because it does complete with that release and reads correctly in the view that holds it. `Next` means in the owner's field of view now, and it has two writers. The owner sets it to mean ready to start, and at most five **open** issues hold it that way; the cap is what keeps the value a decision rather than a copy of everything already accepted. `$finish-change` sets it as well, on the issue its pull request closes, so that work already in flight is legible in the view the owner reads instead of only in the pull request list. Those sit outside the cap: an agent opening a pull request is not choosing what to start next and must never spend one of the five slots that decision uses. A closed issue keeps whatever `Queue` value it had and stops counting, which is why neither kind has to be cleared on merge and why every view that reads `Queue` filters `is:open`.
```
Krzysztof318, line 110: **P3 — Record the `Queue: Parent` exception where the skill reads it from**

**Impact**

`$finish-change` gained "It never overwrites `Queue: Parent`", but *Linking a pull request to its issue* on this page states the write without an exception: do it for every pull request, and nothing else writes the field afterwards. Nothing breaks today, because a `Closes` reference never points at a parent — but the rule now lives in the skill and not in the contract the skill defers to, so the next rewrite from this page loses it.

**Correction**

One sentence under *Linking a pull request to its issue*: `Queue: Parent` is never overwritten, and a `Closes` reference pointing at a parent is the defect to correct.

**Rule** — `docs/operations/issue-tracking.md` line 5: this page is where both skills read the board's rules from.

docs/operations/issue-tracking.md
Krzysztof318: Confirmed against the board and fixed in 3cbda91. `projectV2(number: 4) { workflows }` returns `Code review approved` and `Code changes requested`, so both invented names are gone from the transition list and from the `Ready to merge` paragraph below it, and the board README takes the same correction. `Auto-add sub-issues to project` joins the list as well — it is enabled, and with a second level in the hierarchy it is the workflow that places a child opened under a parent.

.agents/skills/start-task/SKILL.md
Krzysztof318: Right, and the board already works the unconditional way. Fixed in 3cbda91: the step now reads `Parent` where the children span releases, then states separately that every parent's title begins `[P]` whatever its `Queue` value, naming the single-milestone parent as the case that holds an ordinary one.

docs/operations/issue-tracking.md
Krzysztof318: Fixed in 3cbda91. The paragraph says all three fields are missing on arrival and that setting all three is part of opening the issue, and the audit query filters `select(.area == null or .queue == null or .size == null)` and prints `size=` beside the other two. The missing blank line before that paragraph is repaired in the same edit.

CONTRIBUTING.md
Krzysztof318: Already corrected in 3a38e3d, before this review landed — all five places name the `NOT_FOUND` answer with `projectV2: null` rather than an error naming permission. 3cbda91 adds the part that was still missing: `gh` prints the response body, adds `gh: Could not resolve to a ProjectV2 with the number 4` on standard error, and exits `1`, so the line a contributor actually sees is the one that reads least like a permission problem. Verified by running the probe against a project the credential cannot reach.

AGENTS.md
Krzysztof318: Fixed in 3cbda91 — the clause is dropped. It was the last place defining the role by board access, which is exactly what the rest of the change stops doing, and this file has the widest reach of the seven.

.agents/skills/finish-change/SKILL.md
Krzysztof318: Fixed in 3cbda91, with the addition that the board is explicitly not one of the two: the split is the branch name and the remote pushed to, and the board follows the probe `$start-task` ran.

docs/operations/issue-tracking.md
Krzysztof318: Fixed in 3cbda91. One bullet under *Linking a pull request to its issue* states that the write never overwrites `Queue: Parent` and that a `Closes` reference pointing at a parent is a defect in the body rather than a value to write over — noting that this is the one case where not setting `Next` is the correct outcome rather than a failed gate. Agreed on the reasoning: nothing breaks today, but the rule belonged in the contract the skill defers to.

Issue #520, opened 2026-08-07
Apply the Microsoft.Extensions.AI telemetry decorators where the AI boundary first composes a client


Krzysztof318:
Split out of #501, which subscribed the signals the pinned libraries already emit and could deliver every part of that but this one.

`Microsoft.Extensions.AI` does not publish telemetry under a name a host can simply subscribe. Its signals exist only where a client is decorated: `UseOpenTelemetry` on a `ChatClientBuilder` or an `EmbeddingGeneratorBuilder` wraps the client and publishes spans and instruments under `Experimental.Microsoft.Extensions.AI`, with prompt and completion capture behind the decorator's own `EnableSensitiveData` switch. The decorator is therefore applied where a client is composed, and `src/AI` composes none: `AiServiceCollectionExtensions` registers the deterministic chunker and nothing else, `src/AI/Embeddings/`, `src/AI/Orchestration/`, `src/AI/ProviderAdapters/`, and `src/AI/Retrieval/` hold only a `.gitkeep`, and no `IChatClient` or `IEmbeddingGenerator` is constructed anywhere under `src/`.

Applying the decorator now would mean writing a composition helper nothing calls, and subscribing the source name now would collect from a publisher that does not exist. So this waits for the change that first composes a chat client or an embedding generator, and belongs in that change rather than ahead of it — which is what makes it an issue rather than a note.

## User stories

**As a privacy owner**, I want the AI telemetry decorators registered with prompt and completion capture off from the first client the AI boundary composes, so that enabling observability on that boundary never turns a span store into a copy of what was asked and answered.

**As an operator**, I want an AI call to appear in the same trace as the MCP request that caused it, so that a slow answer is attributable to the model call rather than to the code around it.

## Acceptance

- Every `IChatClient` and `IEmbeddingGenerator` the AI boundary composes passes through `UseOpenTelemetry`, applied where the client is built rather than at a call site.
- Sensitive-data capture is explicitly off, and a unit test fails if it is turned on.
- The source and meter name the decorators publish under is subscribed beside the other library names in `TelemetrySubscriptionExtensions`, read from the package's current official documentation rather than from this body.
- Tag cardinality is checked against the same rule the rest of the emitted surface follows: no per-message, per-address, or per-prompt dimension.
- `docs/operations/telemetry.md` states what the AI boundary now emits.
- The unit-test suite and the 85% coverage gate pass.

## Depends on

[...]

Issue #85, opened 2026-07-27
[P] Instrument the pinned libraries so traces, metrics, and logs cover the whole request path


Krzysztof318:
No `specs/` file backs this issue. It records cross-cutting observability work the architecture draft requires — section 22 asks for OpenTelemetry traces across MCP calls, database operations, IMAP synchronization cycles, retrieval, embedding generation, agent runs, and SMTP delivery — that no numbered specification owns.

`src/Host/ServiceDefaultsExtensions.cs` today registers ASP.NET Core, HttpClient, and runtime instrumentation only. Every dependency MailFathom actually spends its time in is invisible: Npgsql and EF Core, MailKit, the Polly pipelines specification 03 introduces, the MCP SDK, and `Microsoft.Extensions.AI`. MailFathom also has no convention for its own `ActivitySource` and `Meter` names, so each feature would invent one. #67 covers the PostgreSQL half through the Aspire integration; this issue owns the rest and the shared foundation both sides register into.

What is known now, to be re-verified against current official documentation when the work starts:

- **Npgsql and EF Core** — connection, command, and pool instrumentation arrives with the Aspire PostgreSQL EF Core integration in #67. This issue owns whatever that leaves unregistered, including the `Microsoft.EntityFrameworkCore` meter.
- **Polly** — `Polly.Extensions` emits metrics under the `Polly` meter plus resilience-event logs, configured through `TelemetryOptions`. Specification 03 builds the pipelines; this issue registers the meter and decides whether enrichment is worth the tag cardinality.
- **MCP SDK** — the C# SDK records histograms under the `Experimental.ModelContextProtocol` meter, such as `mcp.server.session.duration`, and tags activities with `mcp.method.name`, `mcp.protocol.version`, `jsonrpc.request.id`, and `gen_ai.tool.name`. Specification 16 hosts the server; this issue subscribes to the source and meter.
- **MailKit** — no OpenTelemetry instrumentation package exists, so mailbox sessions and synchronization cycles need first-party spans and metrics. MailKit protocol logging must never be routed into telemetry.
- **`Microsoft.Extensions.AI`** — telemetry comes from its OpenTelemetry decorators, whose sensitive-data capture stays off.

## User stories

**As an operator**, I want one trace to follow an MCP tool call through the application use case, the database query, and any AI call it makes, so that a slow `search_emails` is diagnosable from the exported spans rather than by attaching a profiler to the process.

[...]

PR #521, opened 2026-08-07
[#501] Collect the signals the pinned libraries already emit

### LINKED ISSUE #501: Collect the signals the pinned libraries already emit
Sub-issue of #85.

Several pinned libraries already emit spans and instruments that nothing here collects. The MCP SDK records histograms under the `Experimental.ModelContextProtocol` meter — `mcp.server.session.duration` among them — and tags its activities with `mcp.method.name`, `mcp.protocol.version`, `jsonrpc.request.id`, and `gen_ai.tool.name`; `src/Host/ServiceDefaultsExtensions.cs` subscribes neither the meter nor the source, so the MCP surface is invisible in traces and metrics even though the SDK is paying to produce both. `Microsoft.Extensions.AI` exposes its telemetry through decorators that `src/AI/AiServiceCollectionExtensions.cs` does not apply, so nothing the AI boundary does is observable either.

This is the cheapest half of #85: the signals exist and the work is subscribing to them correctly, with the sensitive-data capture the AI decorators offer left off. It also settles what the Aspire Npgsql EF Core enrichment in `src/Infrastructure/HostApplicationBuilderExtensions.cs` already registers, so a meter is not subscribed twice and the `Microsoft.EntityFrameworkCore` meter is not assumed to be covered when it is not.

## Scope correction, 2026-08-07

The `Microsoft.Extensions.AI` half of this issue has moved to #520, and both the user story and the acceptance bullet that carried it are struck below rather than left to read as unmet.

That package publishes nothing under a name a host can subscribe: its signals exist only where a client is decorated with `UseOpenTelemetry`, and `src/AI` composes no client at all — `AiServiceCollectionExtensions` registers the deterministic chunker, `Embeddings/`, `Orchestration/`, `ProviderAdapters/`, and `Retrieval/` hold only a `.gitkeep`, and no `IChatClient` or `IEmbeddingGenerator` is constructed anywhere under `src/`. Applying the decorator would mean writing a composition helper nothing calls, and subscribing the name would collect from a publisher that does not exist. So the work belongs to whichever change first composes one, which is what #520 records; #85's delivery table carries the split.

Everything else below stands and is what this issue is closed on.

## User stories

**As an operator**, I want an MCP tool call to appear in traces with the protocol method and tool name the SDK already records, so that a slow call is attributable to a tool before anyone instruments the code behind it.

**As an operator**, I want the database, resilience, and protocol instruments to arrive in one dashboard with no gap between them, so that a request whose spans stop at the transport boundary reads as a missing subscription rather than as work that did not happen.

~~**As a privacy owner**, I want the AI telemetry decorators registered with prompt and completion capture off, so that enabling observability on the AI boundary never turns a span store into a copy of what was asked and answered.~~ → #520

## Acceptance

- The MCP SDK's meter and activity source are subscribed, with the names read from the SDK's current official documentation rather than from this issue body, which records them as they were when it was written.
- ~~The `Microsoft.Extensions.AI` OpenTelemetry decorators are applied where the AI boundary composes its clients, with sensitive-data capture explicitly off and a test that fails if it is turned on.~~ → #520
- What the Aspire Npgsql EF Core enrichment registers is verified against its current documentation, and whatever it leaves unregistered — the `Microsoft.EntityFrameworkCore` meter among the candidates — is registered here. No meter is subscribed twice.
- Every instrumentation package added is checked for .NET 10 compatibility, current official documentation, and license before it is pinned; versions are pinned centrally in `Directory.Packages.props` and the lock files are regenerated in the same change.
- `THIRD_PARTY_LICENSES.md` records every package this change adds. (#85 says `LICENSES.md`; the file has since been renamed.)
- Unit tests assert the subscribed set, so a meter dropped from the registration fails the suite rather than going quiet in production.
- `docs/operations/telemetry.md` states which libraries' signals the process now collects.
- The unit-test suite and the 85% coverage gate pass.

## Depends on

- Nothing strictly, but land it after #500 so the subscriptions are added in the one place that owns them.

Krzysztof318:
Closes #501

Sub-issue of #85.

Three of the libraries this repository already pins pay to emit telemetry that nothing collected. The MCP SDK spans every JSON-RPC request and records session and per-operation durations; EF Core publishes a meter carrying active contexts, queries, save operations, compiled-query cache hits and misses, execution-strategy failures, and optimistic-concurrency failures. An unsubscribed signal is not a slow signal — it does not exist, and it says so nowhere. The visible cost was an MCP request reading as a gap in the trace between the ASP.NET Core span that carried it and the database commands it issued, and concurrency conflicts observable only one exception at a time.

## What changed

`TelemetrySubscriptionExtensions` becomes the one place the host states what it collects, first-party and library alike. Polly's meter moves into it from `ServiceDefaultsExtensions`, joined by two names that were not collected before:

| Name | Registry | What it buys |
| --- | --- | --- |
| `Experimental.ModelContextProtocol` | activity source **and** meter | MCP spans tagged with the protocol method, negotiated version, transport, session id, JSON-RPC request id, and tool name; session and per-operation duration histograms |
| `Microsoft.EntityFrameworkCore` | meter | Active contexts, queries, save operations, compiled-query cache hits and misses, execution-strategy failures, optimistic-concurrency failures |

**`Npgsql` stays out deliberately.** The acceptance asked that nothing be subscribed twice, and it settles as an absence rather than an addition: the Aspire PostgreSQL enrichment in `AddDatabaseHealthAndTelemetry` already subscribes both the `Npgsql` meter and the `Npgsql` activity source. Verified against `AspireEFPostgreSqlExtensions.cs` and `NpgsqlCommon.cs` at Aspire `v13.4.6` — `AddNpgsql()` for tracing, `AddMeter("Npgsql")` for metrics, and nothing else. A unit test asserts the absence so it cannot be "helpfully" added back.

**EF Core gets no activity source.** It reports through `DiagnosticSource`, not an `ActivitySource`, so no name reaches it; the only route is `OpenTelemetry.Instrumentation.EntityFrameworkCore`, which is prerelease and would span the same database commands the `Npgsql` source already spans.

## How the names were verified

Read from each library's own source at the pinned version, not from the issue body:

- `ModelContextProtocol.Core` v2.0.0 — `src/ModelContextProtocol.Core/Diagnostics.cs` declares one string, `Experimental.ModelContextProtocol`, for both the `ActivitySource` and the `Meter`.
- EF Core v10.0.10 — `EntityFrameworkMetrics.MeterName` is `Microsoft.EntityFrameworkCore`.

Two of the three names are additionally asserted **against the library's own declaration** in the tests, so a rename arriving with a package bump fails the suite rather than quietly emptying a dashboard. The MCP one goes through reflection because the declaration is internal to the SDK and the name is marked experimental by the package that owns it; the EF Core one reads the declaration directly, under an `EF1001` suppression narrowed to that single assertion, since the warning is precisely the hazard being asserted.

## Cardinality and privacy

Every tag on the new instruments is a bounded set: a protocol method, a transport kind, a negotiated version, one of MailFathom's three tool names, an outcome. The high-cardinality identifiers the SDK records — `mcp.session.id`, `jsonrpc.request.id` — go to `activity.AddTag` only, never into the metric `TagList`. The SDK does tag a metric with `mcp.resource.uri`, which would be neither bounded nor free of personal data, but only for the protocol's resource methods; MailFathom's server publishes tools alone, no resources and no prompts, so the tag cannot arise today. Documented so a later resource capability meets the question.

## What is not in this change

The issue's `Microsoft.Extensions.AI` bullet has no target in the tree. That package publishes nothing under a subscribable name: its signals exist only where a client is decorated with `UseOpenTelemetry`, and `src/AI` composes no client at all — `AiServiceCollectionExtensions` registers the deterministic chunker, `Embeddings/`, `Orchestration/`, `ProviderAdapters/`, and `Retrieval/` hold only a `.gitkeep`, and no `IChatClient` or `IEmbeddingGenerator` is constructed anywhere under `src/`. Applying the decorator would mean writing a composition helper nothing calls; subscribing the name would collect from a publisher that does not exist.

So it moves to #520, linked under #85, with #85's delivery table updated to record the split. It belongs to whichever change first composes a chat client or an embedding generator, and it carries no milestone for that reason.

#501's own body now carries a dated *Scope correction* section as well, with the user story and the acceptance bullet that held this work struck through and pointed at #520 — struck rather than deleted, so what moved stays legible. Every acceptance item still standing on #501 is delivered here, which is what makes `Closes #501` accurate rather than a keyword closing an issue over an outstanding item.

## Contracts

No break. No configuration key, database column, MCP tool argument, or deployment contract moves. No package is added, upgraded, or removed, so no pin moves, no lock file is regenerated, and `THIRD_PARTY_LICENSES.md` is unchanged — every name subscribed comes from a package the repository already pins.

## Verification

- `scripts/verify-full.sh` — green. 2867 unit tests, 0 failed; aggregate line coverage 94.73% against the 85% gate.
- `scripts/test-agent-workflow.sh` — 111 passed, 0 failed.
- 12 tests over `TelemetrySubscriptionExtensions`, including the two that read a library's own declaration and the one that asserts `Npgsql` is not subscribed here.

`docs/operations/telemetry.md` now states which libraries' signals the process collects and which registers each; its `describes:` marker gained `src/Infrastructure/HostApplicationBuilderExtensions.cs`, because the page now describes what that file registers. `docs/architecture/outbound-resilience.md` named `ServiceDefaultsExtensions` as where Polly's meter is subscribed and now names the type that holds it.

### DIFF
--- a/docs/architecture/outbound-resilience.md
+++ b/docs/architecture/outbound-resilience.md
@@ -159,11 +159,12 @@ is the layer rather than something MailFathom re-implements:
 
 Polly's metrics stay on and carry the dependency class as the pipeline name, the remote instance as the pipeline
 instance, the event, the attempt number, the outcome, and the duration. Emitting them is not exporting them: the host
-subscribes OpenTelemetry to Polly's meter in `ServiceDefaultsExtensions`, without which the instruments would exist and
-nothing would collect them. Its *logging* is replaced: Polly renders the outcome exception in full, and a mail server
-puts the rejected recipient into its error text. `OutboundResilienceEvents` therefore records a retry, a circuit
-opening, and a circuit closing with the dependency class, the instance, the operation, the failure's type name, the
-attempt number, and the delay — never a message, an address, an identifier, or a payload.
+subscribes OpenTelemetry to Polly's meter in `TelemetrySubscriptionExtensions`, beside every other meter a library
+publishes under its own name, without which the instruments would exist and nothing would collect them. Its *logging*
+is replaced: Polly renders the outcome exception in full, and a mail server puts the rejected recipient into its error
+text. `OutboundResilienceEvents` therefore records a retry, a circuit opening, and a circuit closing with the dependency
+class, the instance, the operation, the failure's type name, the attempt number, and the delay — never a message, an
+address, an identifier, or a payload.
 
 The instance is the configured account identifier and the operation is the folder alias, or the fixed name
 `folder-discovery` for a connection that pins no folder, both carried into the callbacks by the pipeline key and by
--- a/docs/operations/telemetry.md
+++ b/docs/operations/telemetry.md
@@ -1,6 +1,6 @@
 # Telemetry and the Aspire dashboard
 
-<!-- describes: src/Common/Observability/**, src/Host/Observability/**, src/Host/ServiceDefaultsExtensions.cs, src/Infrastructure/Observability/**, src/AppHost/** -->
+<!-- describes: src/Common/Observability/**, src/Host/Observability/**, src/Host/ServiceDefaultsExtensions.cs, src/Infrastructure/Observability/**, src/Infrastructure/HostApplicationBuilderExtensions.cs, src/AppHost/** -->
 
 The host instruments itself with OpenTelemetry throughout — logs, metrics, and traces — and exports none of it unless
 the environment names a destination. Today exactly one environment does that out of the box: a local run under the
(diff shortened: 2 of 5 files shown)


Krzysztof318:
**P2 — Record #501's unmet AI-decorator acceptance item before closing it — fixed, and the finding was right about the half that was missing.**

Taking the second of the two corrections offered: split #501 itself, the way #85's delivery table was updated. The split had been made everywhere except the one place a reader of the closed issue actually looks — #501's own body — which is exactly the gap the rule names. #520 existed, #85's table recorded it, and #501 still listed the acceptance bullet as outstanding, so `Closes #501` would have closed an issue whose own text said it was not done.

[#501's body](https://github.com/Krzysztof318/MailFathom/issues/501) now carries a dated *Scope correction* section, and both the user story and the acceptance bullet that carried the `Microsoft.Extensions.AI` work are struck through with an arrow to #520 rather than deleted — so the history of what moved stays legible instead of the issue quietly appearing to have been smaller than it was. The section also states why the item is not deliverable here: the package publishes nothing under a subscribable name, its signals exist only where a client is decorated, and `src/AI` composes no `IChatClient` and no `IEmbeddingGenerator` at all.

With that, every acceptance item still standing on #501 is delivered by this change, so `Closes #501` is accurate and stays.

The other reading — dropping the keyword and leaving #501 open — would have left an issue with nothing to do on it, since its remaining work is zero once the AI item lives on #520. Reopening after merge would have the same problem plus a manual step nobody is prompted to take.

No code changed, so the head commit is unchanged at `bcbc64f`.

PR #529, opened 2026-08-07
[#427] Add the embedding generator port and its first provider adapter

### LINKED ISSUE #427: Add the embedding generator port and its first provider adapter
## Context

Nothing in MailFathom calls an AI provider today. This adds the first such call and the boundary that contains it: an application-owned port for turning text into vectors, and one adapter behind it in `src/AI/ProviderAdapters`, which is currently an empty directory.

The port is provider-neutral by construction. `Application` and `Domain` see text in and vectors out, and never an SDK type, a model name as a compile-time constant, or a provider exception. ADR 0006 decides where the adapter reads its settings from: configuration declares the provider, the model, the fallback chain, the endpoint, the credential, and the vector-width decision, and the active profile is the immutable record of the identity an activation took from that declaration. The adapter honours what is declared or fails, and a model name is never a compile-time constant.

Two providers are in scope and no more: OpenAI and Azure OpenAI. Both are reached through `IEmbeddingGenerator<string, Embedding<float>>` from `Microsoft.Extensions.AI`, which is the abstraction Microsoft Agent Framework itself moved to and which `AzureOpenAIClient` and `OpenAIClient` both project onto, so the adapter composes an SDK client rather than writing a protocol. That choice is deliberate beyond embeddings: the same framework and the same client construction serve the chat model the answering feature will need, so the provider wiring is built once. Anything beyond those two is #479 and is `Parked`.

Azure OpenAI authenticates one of two ways and the deployment picks which: an API key carried as a `SecretReference`, or a Microsoft Entra credential where no secret exists to provision. The credential is non-interactive by construction — `ManagedIdentityCredential`, `WorkloadIdentityCredential`, `ClientSecretCredential`, `ClientCertificateCredential` — because this is a background service with no user at a keyboard. `InteractiveBrowserCredential` and `DeviceCodeCredential` have no place here, and neither do the developer-tool credentials that `DefaultAzureCredential` chains in, which is why the chain is composed explicitly rather than taken from that default.

Everything the repository already requires of outbound I/O applies here and is the reason this is its own unit of work rather than part of the worker that will use it: a timeout, a named resilience pipeline from specification 03, bounded batch sizes, and a failure classification that separates a caller cancellation from a shutdown, a timeout, an authentication failure, a rate limit, and a transient transport fault. A rate limit answered with an immediate retry is how a provider account gets throttled harder, so classification is the deliverable rather than a detail of it.

## User stories

**As the operator of my own instance** I want a provider outage to be a classified, bounded failure, **so that** synchronization and lexical search keep working while embeddings are unavailable.

**As a maintainer** I want the provider SDK confined to one adapter, **so that** a second provider is a new adapter rather than a change to application code.

**As the operator of my own instance** I want my provider credential to stay out of logs and traces, **so that** enabling embeddings does not create a new way to leak it.

## Acceptance

- The port exposes text in and vectors out, and no provider type crosses out of `src/AI/ProviderAdapters`.
- OpenAI and Azure OpenAI are both supported, through `IEmbeddingGenerator<string, Embedding<float>>` and the client construction Microsoft Agent Framework uses, so the chat model that follows reuses this wiring instead of adding a second one.
- Azure OpenAI authenticates by API key carried as a `SecretReference`, or by an explicitly composed non-interactive Entra credential chain. No interactive or developer-tool credential is reachable from the service, and `DefaultAzureCredential` is not used, because its chain contains both.
- An authentication failure is distinguishable from an unreachable endpoint and from a rate limit, for each of the two providers and each of the two Azure authentication shapes.
- The model, the provider, the endpoint, the credential, and the fallback chain come from configuration and are never compile-time constants. Validation refuses a chain whose pairs differ in model, dimension, metric, or input preparation, naming both pairs and the property they differ on, because ADR 0006 makes a fallback another route to one vector space rather than a second one.
- Falling through the chain is a classified, observable event and changes nothing about what is written: the vectors a fallback endpoint returns belong to the same profile, because its identity is the same.
- Where the provider can return a shortened vector natively — OpenAI's third-generation models accept a requested dimension — that is used in preference to shortening a returned one. Where the adapter must shorten, it renormalizes, and `AllowTrimVectors` being off makes an over-wide model a refusal rather than a silent narrowing.
- A vector whose length does not match the dimension the active profile records is a failure at the adapter, named as such, rather than a row the database rejects later.
- Every call carries an explicit timeout and runs through a named resilience pipeline.
- Failures are classified into cancellation, shutdown, timeout, authentication failure, rate limit, and transient transport fault, and only the safely repeatable ones are retried, with bounded attempts and jittered backoff.
- Batch size and per-request input size are bounded before the provider sees them.
- No mail text, vector, prompt, credential, or provider payload reaches a log, a span, or an exporter.
- The package the adapter uses is pinned centrally, its lock files regenerated in the same change, and recorded in `THIRD_PARTY_LICENSES.md` with its license and, separately, the hosted service's own terms.
- Unit tests cover each failure classification without a network.
- A deterministic in-repository generator implements the port, so everything downstream — the schema, the worker, the backfill, the generation switch — is provable at zero provider cost.
- The tests that exercise the real provider are skipped by default through xUnit v3's `SkipUnless` on a static property reading an environment variable nothing sets, so an ordinary local run and an ordinary pipeline run spend nothing. ADR 0006 records why.
- The `Integration tests` workflow gains an input, defaulting to off, that enables them and supplies the credential. Turning it on without a credential configured fails the run rather than skipping the tests, because a run explicitly asked for that quietly proves nothing is worse than one that never started.
- There are one or two such tests, not a suite: they prove the protocol, the authentication, the failure classification against real responses, and the returned dimension. Everything else belongs to the free path above.
- `tests/AGENTS.md` records the default-skipped rule in the same change, since it governs every test written against a paid provider afterwards.

## Out of scope

Deciding when to embed, which is the worker's question. Storing a vector. Any chat model, though the provider wiring is chosen so that one needs no second mechanism. Any provider beyond the two named above, which is #479.

## Depends on

#424 — ADR 0006 decides what configuration declares, what the profile records, and what a fallback is allowed to be.

Krzysztof318:
Nothing in MailFathom called an AI provider. This adds the first such call and the boundary that contains it: an application-owned port that takes text and answers vectors, and one adapter behind it in `src/AI/ProviderAdapters` reaching OpenAI and Azure OpenAI through `IEmbeddingGenerator<string, Embedding<float>>` from `Microsoft.Extensions.AI`.

Closes #427

## One client construction for both providers

An Azure OpenAI resource's v1 data plane is OpenAI-compatible, so a cloud deployment is the same `EmbeddingClient` pointed at the resource's own `/openai/v1/` address with the deployment's name as the routed model. That is Microsoft's own current guidance, and it is deliberate beyond embeddings: the chat model the answering feature will need is served by this wiring rather than a second one.

`Azure.AI.OpenAI` is not adopted. It would be that second client construction, and it has published no stable release since 2.1.0 — pinning it would have meant a second prerelease pin in a repository that carries exactly one and says why.

## What is declared, and what it refuses

The provider, the model, the width, the metric, the input preparation, the endpoint, and the credential all come from configuration; nothing about a vector space is a compile-time constant. Each entry of the `Embeddings:Endpoints` chain declares its whole geometry, which is what makes a disagreement expressible — and therefore refusable at startup, naming both aliases and the property they differ on. A fallback is another route to one vector space and never a second one.

Two more refusals sit beside it: a declared dimension above what an index covers while `AllowTrimVectors` is off, and an endpoint declaring both a provider key and a Microsoft Entra credential, or neither.

## Authentication, and the chain that is not used

An endpoint carries either a provider key as a `SecretReference`, resolved per request so a rotation needs no restart, or one of four non-interactive Microsoft Entra credentials: managed identity, workload identity, client secret, client certificate.

`DefaultAzureCredential` is refused. Its chain reaches an interactive browser credential — which cannot complete where nobody is at a keyboard — and the developer-tool credentials of whoever last signed in on the host, which would let a deployed service authenticate as an operator's own account. The four shapes are composed explicitly and nothing else is reachable.

## Failure classification is the deliverable

Six classifications: a refused credential, a rate limit, a timeout, a transport fault, a rejected request, and an answer of the wrong shape. Only the three another attempt can change are repeated. A rate limit answered immediately is how an account gets throttled harder; a refused credential repeated buys the same refusal while the account carries the requests.

Cancellation and shutdown stay outside the set, because neither is a statement about the provider — reporting one as a provider failure would let this system's own decision open a circuit against a healthy endpoint.

Falling through the chain answers a different question and now also covers an endpoint the resilience pipeline **declined to call at all**. An open circuit on the first endpoint is precisely the condition a fallback exists for, and ending the request there would make the whole chain unusable for as long as its first entry stayed broken. An answer of the wrong shape does not fall through: every endpoint declares the same geometry, so asking the next one buys a second paid call to learn the same thing.

## One retry layer, and one new port to reach it

The call runs under the existing `AiProviderInvocation` pipeline, keyed per endpoint alias so one unreachable provider does not open the circuit the others are served through. Both the standard HTTP resilience handler and the client library's own retry policy are switched off for it — three layers would have multiplied rather than added.

Reaching that budget from another adapter boundary needed one new application-owned port, `IOutboundOperationRunner`, implemented in `Infrastructure` over the existing executor. The single-layer rule, the failure classification, and the pipeline translation all still happen in exactly one place; the alternative was a second retry mechanism or a project reference from one adapter boundary to another.

## Paying nothing to prove it

An instance that declares no chain embeds nothing, needs no credential, and serves lexical search exactly as before.

A deterministic in-repository generator implements the same port from the text alone — reproducible, of the declared width, of unit length — so the schema, the worker, the backfill, and the generation switch are all provable against a real database at zero provider cost. Its profile names a provider of its own, so a deployment that activated it by accident is visible in the profile row rather than in the quality of its search results.

The two tests that reach a real provider are skipped unless somebody asks: `Integration tests` gains an input defaulting to off. Turning it on without a credential **fails** the run rather than skipping, because a run somebody requested and which then quietly proved nothing is worse than one that never started. `tests/AGENTS.md` records that rule for every test written against a paid provider afterwards.

## Breaking changes

None. `Embeddings` is a new configuration section and an absent one is a supported deployment; no existing key, tool contract, database column, or deployment asset moves.

## Dependencies

Three new central pins, all MIT, all verified against the published nuspec:

- `OpenAI` 2.12.0 and `Microsoft.Extensions.AI.OpenAI` 10.8.1 — the provider client
- `Azure.Identity` 1.21.0 — the four non-interactive credentials

`THIRD_PARTY_LICENSES.md` carries both rows plus the transitive closure, and two new *Hosted services* rows for OpenAI and Azure OpenAI with their data-use and abuse-monitoring retention terms and the operator-side conditions that follow from mail text leaving the deployment. The published-assembly count is corrected from 66/60 to the measured 77/71.

## Verification

`scripts/verify-full.sh` green: 3093 tests, 0 failed, aggregate line coverage 94.79%.

https://claude.ai/code/session_01KRjzrZtEA8ozmnt4bkFCM1


.github/workflows/integration-tests.yml
```diff
@@ -74,6 +83,16 @@ jobs:
       # `dotnet-ef` is fetched by the migration resource itself and needs no install here.
       - name: Run integration tests
         run: bash scripts/run-integration-tests.sh
+        env:
+          # Absent on every run that did not ask, which is what makes the provider-contract tests
+          # skip and cost nothing. The `Release` workflow calls this one without the input, so a
+          # release never spends provider credit either.
+          MAILFATHOM_EMBEDDING_CONTRACT_TESTS: ${{ inputs.run_provider_contract_tests || 'false' }}
+          MAILFATHOM_EMBEDDING_API_KEY: ${{ secrets.EMBEDDING_PROVIDER_API_KEY }}
+          MAILFATHOM_EMBEDDING_ADDRESS: ${{ vars.EMBEDDING_PROVIDER_ADDRESS }}
```
Krzysztof318, line 92: If lack of secrets break the pipeline in normal run with run_provider_contract_tests: false?

tests/IntegrationTests/Embeddings/EmbeddingProviderContractTests.cs
```diff
@@ -0,0 +1,126 @@
...
+        // Assert
+        Assert.Equal(2, vectors.Count);
+        Assert.All(vectors, vector => Assert.Equal(plan.Identity.Dimension, vector.Dimension));
+    }
+
+    /// <summary>
+    /// The one failure worth a paid call to prove: the classification the adapter derives has to match what the
+    /// provider actually answers to a credential it does not accept, and no unit test can establish that.
+    /// </summary>
+    [Fact(SkipUnless = nameof(ProviderContractTestsRequested))]
```
Krzysztof318, line 66: Is not better control this with Collection Fixtures? also keep in mind only generation can run in parallel but anything related to db has to be seqeuntial with any other db tests

.github/workflows/integration-tests.yml
Krzysztof318: No — an ordinary run with the input off skips both tests and reads no secret at all.

Two independent reasons, and I checked both rather than reasoning from the YAML alone:

**The environment.** An undefined `secrets.*` or `vars.*` expands to the empty string; it is not an error and not a workflow-validation failure. So on a run that asked for nothing, `MAILFATHOM_EMBEDDING_API_KEY` is simply empty.

**The predicate.** The skip condition reads *only* the enabling variable, and `Plan()`/`ApiKey()` — the two that would throw on a missing secret — are called from inside the test bodies, which never run. I probed the shipped `MailFathom.IntegrationTests.dll` by reflection rather than trusting the source:

```
(unset)   -> requested=False
""        -> requested=False
false     -> requested=False
False     -> requested=False
true      -> requested=True
True      -> requested=True
yes       -> requested=False
1         -> requested=False
```

Note the last two: only a parseable boolean `true` turns it on, so a typo fails safe. `null` and `""` are handled identically by the `is { Length: > 0 }` pattern, which matters because .NET is inconsistent about which one an externally empty variable produces.

The asymmetry you would want is intact in the other direction: with the input **on** and a secret missing, `Required()` throws and the test *fails* — it does not skip.

What was missing was any proof of this, which is what your question exposed. A later edit to the input default, to the `|| 'false'` fallback, or a `run_provider_contract_tests` input added to `workflow_call` would each start spending provider credit on an ordinary run with nothing failing — and `release.yml` reaches this suite through `workflow_call`, so the third one would put it in every release. Frozen in `a192be8` as a workflow contract test beside the other workflow invariants, and I verified it by flipping each of the three properties in turn and confirming it fails on each.

tests/IntegrationTests/Embeddings/EmbeddingProviderContractTests.cs
Krzysztof318: A collection fixture would be worse here, and the reason is precisely your other comment.

xUnit v3 constructs a collection fixture **even when every test in the collection is skipped by `SkipUnless`**. I probed it rather than assuming — a throwaway xUnit v3 3.2.2 project, a collection fixture whose constructor throws, two tests skipped by a condition that is always false:

```
PROBE: fixture constructed
  SkippedTestsWithAFixture.SkippedByCondition [FAIL]
    Collection fixture type 'ProbeFixture' threw in its constructor
  SkippedTestsWithAFixture.AlsoSkippedByCondition [FAIL]
Total: 2, Failed: 2, Skipped: 0
```

So resolving the plan or the credential in a fixture turns an ordinary run with no secrets from *two skipped* into *two failed* — exactly the pipeline break you asked about on the workflow. Keeping the reads inside the test bodies is what makes the skip cost nothing, and it is the only shape that does.

On parallelism — agreed on the rule, and this class is on the right side of it. It touches no database and no mailbox: it composes an `IHttpClientFactory` and calls the provider over HTTPS, nothing else. It consumes no fixture, so it cannot reach the orchestrated database even accidentally, which is why it joins no collection — the class comment says so and that is what it is for.

It also cannot disturb the measurement-sensitive tests. All three collections disable parallelization, and `tests/AGENTS.md` records that xUnit runs those after the parallel-capable ones, so this class is finished before the rate-limit collection starts.

Nothing to change here, so no code moved for this one. If you would rather it were serialized anyway — belt and braces, or because you expect it to grow into something that does touch the database — say so and I will put it in `OrchestratedInfrastructureCollectionDefinition`; the cost is that these tests would then hold the sequential phase open for the length of a provider round trip.

PR #540, opened 2026-08-07
[#536] Retry a provider-contract test a rate limit failed, instead of calling it a defect

### LINKED ISSUE #536: Retry a provider-contract test that a rate limit or an overload failed, instead of reporting it as a defect
## Context

**No numbered specification backs this issue.** `tests/IntegrationTests/Embeddings/EmbeddingProviderContractTests.cs` is
the one place in the repository where a test calls a real, paid provider over the internet. Both of its tests are
deliberately written without a resilience pipeline around them — `PassThroughOutboundOperationRunner` is passed to the
generator so that what the test proves is the adapter's own behavior rather than the pipeline's — which means every
transport answer the provider gives reaches the assertion unfiltered.

That is the right shape for what these tests prove, and it is also what makes them fragile in a way no assertion in them
is about. Hosted model providers answer `429 Too Many Requests` when a per-key or per-tenant rate limit is momentarily
exceeded and `529` when the service itself is overloaded, and both are transient by construction: the same request a few
seconds later succeeds. Neither says anything about whether the adapter speaks the protocol, authenticates, classifies a
refusal, or returns the declared width — the four things these tests exist to establish. So a `429` today reports a red
`Integration tests` run whose only cause is that somebody else was calling the same endpoint, and the only remedy
available is to dispatch the whole workflow again, which restarts a PostgreSQL container and the entire suite to re-run
two tests.

`tests/AGENTS.md` § *How many, and in what order* already names this cost — "it costs them again on every rerun of a
flake" — and the suite has no mechanism that answers it. Retrying at the test-case level is the mechanism: `xRetry.v3`
publishes `[RetryFact]` for xUnit v3, which re-runs a failed test case up to a bounded number of attempts and reports
only the final outcome. The `xRetry.v3` package targeting xUnit v3 is at `1.0.0-rc3`; the xUnit v2 line it was split
from has been stable since 2020, the package is MIT, and the dependency is test-only and never redistributed.

The tension worth stating rather than hiding: a retry attribute cannot tell a rate limit from a wrong answer, so it will
also re-run a genuine assertion failure before reporting it. That is acceptable here for reasons specific to this class
which do not generalise — the failure is still reported, only later; the attempt count is bounded; the class holds
exactly two tests, so the worst case is a handful of calls over two short strings; and the alternative already in place
is re-dispatching the whole suite by hand, which costs strictly more. It is not acceptable anywhere the suite talks to
infrastructure it controls, where a failure is a defect rather than weather.

## User stories

**As the owner dispatching `Integration tests` with the provider-contract input on**, I want a run that hit a rate limit
to retry the affected test rather than fail, so that a red result means the adapter is wrong rather than that the
provider was busy.

**As somebody reading a failed provider-contract run**, I want the failure to have survived several attempts before it
was reported, so that the first thing I do is read the assertion rather than dispatch the workflow again to see whether
it repeats.

**As a future author of a provider-contract test for a chat or LLM adapter**, I want `tests/AGENTS.md` to say that a test
calling a real provider carries the retry attribute, so that the rule is found where the test is written rather than
rediscovered from a flake.

## Acceptance

- `xRetry.v3` is pinned centrally in `Directory.Packages.props` and referenced by `tests/IntegrationTests` only. That
  project carries no `packages.lock.json` by design, so none is added.
- Both tests in `EmbeddingProviderContractTests` carry `[RetryFact]` with at most three attempts and a delay between
  them, and keep their `SkipUnless` condition, so a run nobody asked for still costs nothing and still skips rather
  than retrying.
- The attempt count and the delay are named once, as constants, rather than repeated as literals on each attribute.
- The class documentation says why the retry is there and what it deliberately does not distinguish, so the trade-off
  above is readable in the file rather than only in this issue.
- `tests/AGENTS.md` states the rule for a test that reaches a real provider: bounded retries at the test-case level, and
  why that licence stops at the provider boundary and does not reach a test against the orchestrated database or
  mailbox.
- `THIRD_PARTY_LICENSES.md` carries `xRetry.v3` in the test-only section, with its license expression, its upstream
  repository, and the note that it is not redistributed.
- No production code changes, and the pipeline's resilience policy is untouched: this is about how a test reports a
  transient provider answer, not about how the running service handles one.

Krzysztof318:
Closes #536

## What moved

The two tests in `EmbeddingProviderContractTests` are the only ones in this repository that call a real, paid provider
over the internet, and they are deliberately composed without a resilience pipeline — `PassThroughOutboundOperationRunner`
is handed to the generator so what the test proves is the adapter's own behavior rather than the pipeline's. That leaves
every transport answer reaching the assertion unfiltered, including the two that say nothing about the adapter at all: a
`429` when a key or a tenant briefly exceeds its rate limit, a `529` when the service is overloaded. Both are transient
by construction, and the only remedy available after one was to dispatch the whole `Integration tests` workflow again —
restarting a PostgreSQL container and the entire suite to re-run two tests.

Both tests now carry `[RetryFact]` from `xRetry.v3` in place of `[Fact]`, over three attempts with five seconds between
them, and keep the `Skip` and `SkipUnless` pair #539 established unchanged, so a run nobody asked for still costs
nothing and still skips rather than retrying.

## The trade-off, stated rather than hidden

A retry attribute cannot tell a rate limit from a wrong answer, so a genuine assertion failure is re-run before it is
reported. That is acceptable here for reasons specific to this class: the failure is still reported, only later; the
attempt count is bounded; the class holds exactly two tests over two short strings; and the alternative already in place
costs strictly more. It is not acceptable anywhere else in this suite, and `tests/AGENTS.md` now says where the licence
stops — a test against the orchestrated database, the orchestrated mailbox, or a composed host is talking to
infrastructure the run itself started, where a failure is a defect and retrying one would hide it.

The attribute also sits above the assertion rather than under it, which is what keeps it from being the resilience
pipeline arriving by another route: a passing test still makes exactly one call, and the credential refusal is still
classified from the provider's first answer.

## Dependency

`xRetry.v3` `1.0.0-rc3`, pinned centrally and referenced by `tests/IntegrationTests` alone. It is a release candidate
deliberately — it is the only line of the package that supports xUnit v3, the xUnit v2 line it was split from has
shipped stable since 2020, and the risk is bounded by the dependency being test-only and reachable from two tests. MIT,
verified against the package's own `xRetry.v3.nuspec` at tag `xRetry.v3_v1.0.0-rc3` and against GitHub's license
detection on `JoshKeegan/xRetry`; recorded in `THIRD_PARTY_LICENSES.md` under *Test-only*. Its one dependency,
`xunit.v3.extensibility.core [3.0.0,4.0.0)`, is already registered at 3.2.2 as a transitive of `xunit.v3.mtp-v2`, so
the transitive closure does not widen. `tests/IntegrationTests` carries no `packages.lock.json` by design and none is
added.

## Breaking changes

None. No production code changes, no configuration key, no database schema, no MCP tool contract, and no deployment
contract moves. The running service's resilience policy is untouched — this is about how a test reports a transient
provider answer, not about how MailFathom handles one.

## Verification

- `scripts/verify-full.sh` on this head — green. `failed: 0`, aggregate line coverage 94.44% (8863/9385) against the
  85% gate, `dotnet format --verify-no-changes` clean across 1062 files, the agent-workflow contract suite passing.
- The suite's assembly fixture starts a PostgreSQL container for any run, so neither contract test could be executed
  locally; this repository runs the integration suite in the pipeline rather than on a developer machine. **The retry is
  first exercised against a provider by an `Integration tests` dispatch with `run_provider_contract_tests` on**, which
  is the owner's call to make.
- What that leaves unproven was proven separately, because #539 had just shown that a conditional skip is exactly where
  this class breaks quietly. A throwaway xUnit v3 project outside the repository, referencing the same two pinned
  packages, established three things and was then deleted: `[RetryFact]` carrying `Skip` and a false `SkipUnless`
  reports `[SKIP]` with its message and `Failed: 0`, so the retry attribute does not reintroduce the defect #539 fixed;
  a test that first passes on its third call passes, so the retry actually fires; and a test that never passes reports
  its failure at `attempt 3`, so three is the total number of attempts rather than one plus three retries. The source
  agrees — `RetryFactDiscoverer` reads the skip through `TestIntrospectionHelper.GetTestCaseDetails` and forwards
  `SkipReason` and `SkipUnless` unchanged into `XunitTestCase`'s base constructor — but the executable check is what the
  claim rests on.
- Discovery against the real project finds 75 tests including both contract tests, with no `ExecutionErrorTestCase` in
  their place.

## One correction outside the diff's subject

`docs/operations/local-development.md` § *Package sources and lock files* stated 48 direct pins in
`Directory.Packages.props`. The file held 51 before this change and 52 after, so the number was already stale by three;
it is corrected here because this change is the one that moves it. The neighbouring claims were checked rather than
assumed: 18 lock files across 20 projects, both still accurate.


tests/IntegrationTests/IntegrationTests.csproj
```diff
@@ -8,6 +8,10 @@
     <PackageReference Include="MailKit" />
     <PackageReference Include="NSubstitute" />
     <PackageReference Include="MimeKit" />
+    <!-- Bounded test-case retries, for the two tests that call a real model provider over the internet and therefore
```
Krzysztof318: comment unnecessary

Krzysztof318: Removed. The two things it said are now said where they belong — `tests/AGENTS.md` carries the rule and where it stops, and the test class carries why these two tests are the ones retried — so the project file is back to naming the reference and nothing else, like every line around it.

PR #616, opened 2026-08-08
[#613] Prove against real infrastructure what only it can prove, and let correctness outweigh the suite's cost

### LINKED ISSUE #613: Prove against real infrastructure what only it can prove, and let correctness outweigh the suite's cost
The unit suite proves what a substitute can prove, and `[RequiresIntegrationCoverage]` marks the sixty places where that is structurally not enough. An audit of those markers against what `tests/IntegrationTests` actually reaches found the gaps below. Each one is a failure that cannot surface in a unit test and does not surface as an exception in production either: it surfaces as a wrong answer, a budget that stops counting, or a query that quietly stops using its index.

The suite's own rule is what has kept them open. `tests/AGENTS.md` § *How many, and in what order* opens by pricing an integration test in container start-up, real I/O, and continuous-integration minutes, and asks for the smallest set that proves what only real infrastructure can prove. The instruction here reverses which side of that trade is the tie-breaker: the program behaving correctly outweighs what the suite costs to run. The rule is rewritten to say so, and everything it protected that was never about cost — no unit-testable logic in this suite, one collection per shared resource, ordering only where a claim is about a second run — is kept.

What is **not** relaxed is ADR 0006's default that a paid provider call never happens unless somebody asks for one. Every test added here runs on the free path: a real database, a real mailbox, and the in-repository deterministic generator. The `run_ai_provider_contract_tests` switch and the two adapters behind it are untouched.

## User stories

- As the maintainer, I want the embedding spend ceiling proved against two concurrent transactions, so that a ledger that silently stops counting is found by the suite rather than by an invoice.
- As the maintainer, I want the hybrid ranking that `search_emails` and `ask_mail` both read proved over one real database, so that a mismatch between the lexical and the vector halves cannot rank mail wrongly while both halves pass their own tests.
- As a contributor reading `tests/AGENTS.md`, I want the rule to state that correctness outweighs the suite's runtime, so that adding the test a risk deserves is not an argument against the contract.

## Acceptance

- `EmbeddingSpendLedger` carries an integration test in which two concurrent sessions each charge one period and the ledger reads back their sum, proving the `ON CONFLICT ... DO UPDATE` adds rather than overwrites.
- `MailboxSearchReader` carries an integration test over the real database in which both halves — the full-text index and the pgvector index — match the same seeded mail, and reciprocal-rank fusion returns each message once, ranked, with the identity both halves agree on.
- The chain from chunking through `DeterministicTextEmbeddingGenerator` and pgvector to the retrieval an answer is built from is proved end to end on the free path, including that a citation names mail that is actually stored.
- `MailboxMutationReconciliationStore` carries an integration test; it is one of the two marked ports the suite reaches by neither name nor port today.
- The MCP tool contract is proved over the composed host: a tool call over the real transport returns the documented shape against seeded mail.
- `tests/AGENTS.md` § *How many, and in what order* states that proving the program correct outweighs pipeline time, and keeps the unit-testable-logic rule, the shared-resource collection rule, and the ordering rules unchanged.
- Every added test runs on the free path and none is gated on `AiProviderContractRun`.

## Correction to the audit

One gap this issue was opened with is withdrawn. The audit read the absence of an index-usage test as a gap; `EmailVectorSearchIndexReader` records the opposite decision in its own remarks — the structured filters join the ranking, an HNSW scan orders the whole table and cannot carry a filter on a joined table, so the ranking is deliberately exact and deterministic because the fused order rests on that determinism. A test asserting the index serves that query would assert against a decision the class states, so it is not written and the acceptance item for it is struck. Should a measurement ever argue for an approximate ranking, it belongs in that method and in an issue of its own.

Krzysztof318:
Closes #613

The `[RequiresIntegrationCoverage]` marker names sixty places whose verification can only live in the integration suite. Read against what that suite actually reaches, six of them were reached by nothing, and each one fails the same way: silently. A spend total that stops adding, two halves of a ranking that stop agreeing on which message they are ranking, a reconciliation read that recognizes none of its own rows, a tool contract nothing exercises over the transport it is published on.

## The tests

Five, and every one of them runs on the free path — a real database, a real mailbox, and the in-repository deterministic generator — so nothing here spends a provider call.

- **`OrchestratedEmbeddingSpendLedgerTests`.** Two sessions charge one period at the same time and the period ends up holding both. The ledger is one raw `INSERT ... ON CONFLICT DO UPDATE` that never executes in a unit test, so a conflict target on the wrong column or an assignment that replaced the total instead of adding to it would pass there and reach an operator as a spend ceiling that quietly stopped counting. The two charges are dispatched together rather than awaited in turn, which is a requirement: the first statement holds the row until its transaction ends.
- **`OrchestratedHybridRetrievalTests`.** Both halves of the ranking already had tests; the fusion over one real database had none. The message carrying the query's word arrives from the full-text index, a message sharing no word with the query arrives from pgvector, a message with neither is reached by neither, and each arrives once — which is the claim fusion rests on, since the two halves are separate queries over separate indexes producing separate row shapes. The lexical ranking is run alone beside it, so "this one came from the other half" is a fact rather than an assumption. The second test follows the same window to the passages an answer is built from and establishes that every identifier a model would be asked to cite resolves to a row this database holds.
- **`OrchestratedMutationReconciliationReadTests`.** The whole convergence loop for a relocation MailFathom performed: the run meeting the new occurrence finds the record that explains it, the run meeting the source folder's gap finds the same record, and once the placement is observed the read stops returning it. A second test proves a `\Seen` this instance set is told apart from one a person set. Each read narrows on a batch of UIDs, a mutation name held as text, a stage, and a nullable UID column that must match a value and never a null; every way that can fail is silent in the same direction — a run stops recognizing its own work and reconciles it as somebody else's change.
- **`ComposedMcpToolContractTests`.** A `search_emails` call over the composed host's MCP endpoint, answered from seeded mail, in the fields the tool's own description tells a client to read. The sibling class proves what the controls in front of the endpoint do; nothing proved what the endpoint answers when they let a call through.

## The rule

`tests/AGENTS.md` § *How many, and in what order* opened by pricing an integration test in container start-up and continuous-integration minutes and asking for the smallest set that proves what only real infrastructure can prove. It now says the program behaving correctly outweighs what the suite costs to run, and that a class with six claims only real infrastructure can settle earns six tests.

What is kept is everything that was never about the cost: a rule a substitute already exercises still belongs in the unit suite, one collection per shared resource, ordering only where a claim is about a second run, and an absence asserted only beside the control that would report it present. What is added is the boundary the relaxation does **not** cross — a test that reaches a real AI provider stays behind the one switch, defaults to spending nothing, and stays at one or two per adapter, because what bounds that is somebody's credit rather than a pipeline's minutes. [ADR 0006](https://github.com/Krzysztof318/MailFathom/blob/main/docs/decisions/0006-embedding-profile-identity-lifecycle-and-activation-cost.md) is untouched.

## What the tests needed to exist

- The harness gains `IMailAccountCatalog`, the port every mailbox read resolves its scope through. Without it a search resolves nothing rather than narrowing to this account.
- The composed host is configured with the one account identifier the suite stores its mail under, and with nothing else about that account: configuration is what defines the accounts a deployment serves and it is read whether or not synchronization runs, so a host naming none answers every mailbox read with an empty window over a database that is not empty. With synchronization off, a server or a credential beside the identifier would be configuration nothing acts on.

## One gap withdrawn

The audit read the absence of an index-usage test as a gap. `EmailVectorSearchIndexReader` records the opposite decision in its own remarks: the structured filters join the ranking, an HNSW scan orders the whole table and cannot carry a filter on a joined table, so the ranking is deliberately exact — and determinism is what the fused order rests on. A test asserting the index serves that query would assert against a decision the class states, so it is not written.

## Verification

`scripts/verify-full.sh` passed: 4195 unit tests, 0 failures, aggregate line coverage 95.11% against the 85% gate. The integration suite itself is dispatched through the `Integration tests` workflow after this push rather than run locally.

## Dispatched runs

The first dispatch found three failures in the new tests, which is the whole argument for the suite in one paragraph: both local gates were green and both were blind to them. `MailboxSearchReader` takes `EmailSearchSnippetBounds`, composed by a composition root from the `MailboxSearch` section rather than registered by `AddInfrastructure`, so the harness could not activate the use case; and a folder alias is upper-cased when it is created, so what the passage and the tool result publish is the canonical form rather than the caller's spelling. Both are fixed in the second commit.

[Run 31277803091](https://github.com/Krzysztof318/MailFathom/actions/runs/31277803091) is green on the branch head: `total="110" executed="106" passed="106" failed="0" notExecuted="4"`, the four not executed being exactly the paid provider contract tests the AI switch skips when it is off.


tests/IntegrationTests/Hosting/ComposedMcpToolContractTests.cs
```diff
@@ -0,0 +1,178 @@
...
+    private const string QueryTerm = "chartering";
+
+    private const string SeededSubject = "mcp-tool-contract-seeded";
+
+    /// <summary>
+    /// The whole path a client takes: a call by the tool's published name, answered from the mail this deployment
+    /// serves, in the fields the tool's own description names.
+    /// </summary>
+    [Fact]
+    public async Task CallTool_SearchingForSeededMail_AnswersInTheShapeTheToolContractPublishes()
```
Krzysztof318, line 59: Remember if anything uses ai should be skipped if pipeiline ai param is false

Krzysztof318: Checked this against what the new tests actually reach, and none of them needs the skip — but the rule you are stating is exactly the one this PR writes down, so here is the evidence rather than an assurance.

**The switch gates a paid provider call, not the AI boundary.** `MAILFATHOM_AI_CONTRACT_TESTS` exists because a call to a hosted provider costs money per unit of mail; ADR 0006 requires the in-repository deterministic generator to exist precisely so that everything downstream of that boundary — the schema, the generation, the backfill, the ranking, the retrieval a model is handed — is provable against a real database at zero provider cost. Gating the free path behind the paid switch would mean those tests never run on an ordinary dispatch, which is the outcome the generator was added to prevent.

**What each new test reaches, checked rather than assumed:**

- `ComposedMcpToolContractTests` — the file this comment sits on — touches no AI at all. `search_emails` is answered lexically here, because the app model gives the composed host no AI configuration whatsoever: `grep -n "Embedding\|Chat\|Provider\|Ai" src/AppHost/Program.cs` returns nothing.
- `OrchestratedHybridRetrievalTests` does use the AI boundary, and only through `DeterministicTextEmbeddingGenerator`, which hashes text in-process and opens no socket. `OrchestratedMailFathomServices` registers `AddDeterministicTextEmbeddings` and no provider adapter at all, so there is no configuration of the switch under which these tests could resolve a provider-backed generator.
- `OrchestratedEmbeddingSpendLedgerTests` writes to the ledger table directly and asks no generator for anything, despite the name.
- `OrchestratedMutationReconciliationReadTests` is database-only.

None of the four references `AiProviderContractRun`, `MAILFATHOM_EMBEDDING_*`, or `MAILFATHOM_CHAT_*`. The two classes that do — `EmbeddingProviderContractTests` and `ChatProviderContractTests` — are untouched by this PR and stay behind the switch with their `SkipUnless` and their retry.

The rule is now stated in `tests/AGENTS.md` § *How many, and in what order*, as the one cost the relaxation deliberately does not touch: a test that reaches a real AI provider stays behind the single switch, defaults to spending nothing, and stays at one or two per adapter, because what bounds it is your credit rather than a pipeline's minutes.

If you did mean the stronger rule — that a test merely *exercising* AI-boundary code should also skip when the parameter is off — say so and I will gate them, but it is worth naming what that costs: the hybrid ranking and the retrieval an answer is built from would then be proved by nothing on an ordinary run, which is the gap this PR was opened to close.

Krzysztof318: The dispatched run settles this with numbers rather than with my reading. [Run 31277803091](https://github.com/Krzysztof318/MailFathom/actions/runs/31277803091), with `run_ai_provider_contract_tests` off, reports `total="110" executed="106" passed="106" failed="0" notExecuted="4"`.

The four the switch skipped are exactly the four that spend money:

- `EmbeddingProviderContractTests.GenerateAsync_AgainstTheRealProvider_AnswersInTheDeclaredSpace`
- `EmbeddingProviderContractTests.GenerateAsync_WithACredentialTheProviderRefuses_IsClassifiedAsSuch`
- `ChatProviderContractTests.AnswerAsync_AgainstTheRealProvider_ReturnsAnAnswerWithinTheDeclaredBudget`
- `ChatProviderContractTests.AnswerAsync_WithACredentialTheProviderRefuses_IsClassifiedAsSuch`

Every test this PR adds ran and passed on the same run, including the two that exercise the AI boundary through the deterministic generator. So the parameter is doing exactly what you are asking it to do, and it is doing it without taking the free path down with it.

PR #624, opened 2026-08-09
[#622] Let a deployment state a reasoning effort, and reach the API surface a reasoning model requires

### LINKED ISSUE #622: Let a deployment state a reasoning effort, and reach the API surface a reasoning model requires
A current reasoning model cannot answer a question here, and the operator finds out from a provider `400` in the middle
of their first `ask_mail` call:

```
ChatGenerationFailedException: Chat endpoint '<alias>' rejected the request itself.
 ---> ClientResultException: HTTP 400 (invalid_request_error)
Parameter: reasoning_effort
Function tools with reasoning_effort are not supported for <model> in /v1/chat/completions.
To use function tools, use /v1/responses or set reasoning_effort to 'none'.
```

The provider names both ways out and this deployment can take neither.

- **`/v1/responses` is unreachable.** The chat path runs through `OpenAI.Chat.ChatClient.CompleteChatAsync`, reached by
  `ProviderChatModelClient` and `ResilientChatClient`. Nothing selects an API surface, so chat completions is not a
  default here — it is the only thing the adapter can do.
- **`reasoning_effort` cannot be stated.** `grep -ri reasoning src/` finds prose and nothing else: MailFathom never
  sets the parameter, and the `Chat` section carries no key for it. `Chat` states `Model`, `Address`,
  `MaxOutputTokens`, `Temperature`, `TopP`, `MaxMessagesPerRequest`, `MaxRequestCharacters`, `RequestTimeout`, and the
  credential, and that is the whole surface.
- **Function tools are not optional.** A run is a tool loop by construction — the model asks for mail, the retrieval
  context provider answers, the model writes. Removing tools removes the capability rather than working around the
  refusal.

So the intersection of "models this deployment can answer with" excludes the current reasoning family, and nothing says
so before the call: startup validates the section, the capability gate reports `Available`, the tool is advertised, and
the first question is where it breaks.

## Stories

- As an operator, I want to state the reasoning effort a run is conducted at, so that a reasoning model can be used
  where the provider requires the parameter to be named.
- As an operator, I want a run to be conducted through the API surface the model supports, so that choosing a model is
  a choice about answers rather than about an HTTP path I cannot see.
- As an operator who names a model this deployment cannot use, I want to be told at startup or in the reference, not by
  a `400` inside a question a client is waiting on.

## Scope

Two halves, and they are separable — the first is worth having even if the second is deferred.

- **`Chat:ReasoningEffort`.** A declared value, sent when it is written and omitted entirely when it is not, exactly as
  `Temperature` and `TopP` already behave — a parameter some models reject outright must not acquire a literal default.
  `none` has to be expressible, because that is what the refusal above asks for.
- **The Responses API.** Whether a run is conducted through `/v1/responses` rather than `/v1/chat/completions`, and
  whether that is a declared choice or one derived from the model, is the design decision this issue carries. Derive it
  and a deployment cannot correct a wrong guess; declare it and every operator has to know something about their model
  that the provider already knows. Azure's OpenAI-compatible surface and any self-hosted OpenAI-compatible server are
  part of that decision, because neither necessarily offers both paths.

## Acceptance

- A reasoning model that refuses the combination above can be configured and answers a question, with the effort stated
  in configuration.
- A section that writes no reasoning effort sends none, and the request is byte-identical to what it is today.
- Whatever the Responses decision turns out to be, it is recorded — an ADR if it changes how a provider is reached, a
  paragraph in `docs/features/chat-generation.md` if it does not.
- `docs/operations/configuration-reference.md` § `Chat` states the new key with its constraint and its `Change` class,
  and states what a model must support for `ask_mail` to work at all.
- A provider refusal of this shape keeps the classification it has today: rejected by the provider, not retried, the
  capability degraded for the recheck window.
- Unit tests cover the option's validation and the request built from it, per `tests/AGENTS.md`.

## What follows this

#623 makes the same section reloadable without a restart, and it waits on this one: the set of keys a reload has to
carry is settled here, so building it first would mean building it twice.

Krzysztof318:
A current reasoning model refuses function tools beside an unstated reasoning effort, and names two ways out: state the effort, or use `/v1/responses`. This deployment could take neither — nothing set `reasoning_effort`, and the chat path reached chat completions and nothing else — so an operator learned it from a provider `400` in the middle of their first `ask_mail` call, after startup had validated the section and the capability gate had reported the provider available.

Both halves of the issue are here.

## `Chat:ReasoningEffort`

The provider's own word for the level, carried through as written: `none`, `minimal`, `low`, `medium`, `high`, `xhigh`, or whatever a later model adds. Sent only when written, exactly as `Temperature` and `TopP` already behave — a model that does not reason rejects the parameter outright, so a literal default would turn every call such a deployment makes into a rejected request. A section that writes none produces a request byte-identical to today's, which is asserted rather than assumed.

**`none` is expressible and is not the same as writing nothing.** It states an effort of none and sends it, which is precisely what the refusal above asks for.

**Not a set fixed at build time.** The first version of this used an enum of five levels and it was the wrong shape: `xhigh` arrived after the levels beneath it, and a set here would make a MailFathom release the price of using the next one. Worse, routing the value through the provider-neutral reasoning member of the AI abstraction — whose set is fixed when *its* package is compiled — already made `minimal` unreachable, a level the provider ships today. So the effort is stated through the client library's own request options, whose level type is built from a string. The two APIs frame it differently, so the hook is chosen per API; everything else on the request is untouched, because the abstraction fills the members it owns over the top and the bounds, sampling parameters, instruction, and tools all still go out.

Startup checks the **shape** and never the vocabulary — one word of letters and digits, single hyphens or underscores between, at most 32 characters. A list held here could only be the list that was true when it was written, so a level the model does not know is a *request refused* rather than something this pre-empts. This is the same choice `Model` makes and for the same reason.

## `Chat:Api` — the design decision the issue carried

`ChatCompletions` or `Responses`, **declared rather than derived**.

Deriving it would mean reading the routed model name, and that name is not a model identity: for a cloud deployment it is whatever the operator called the deployment, so a derivation would be guessing from a string the operator invented — and a wrong guess is one nothing in the deployment could correct. Declaring it costs an operator one line about their own provider, which is a thing they already know. Azure's OpenAI-compatible surface and any self-hosted OpenAI-compatible server are exactly why the question arises: neither necessarily offers both paths, and nothing about a model name says which.

This one **stays a closed enum**, deliberately, and it is not the same kind of thing as the effort: it selects which client this build constructs, so a third value would be code rather than configuration and no model can add one. Choosing `Responses` against a server that does not serve that path is a *request refused* — not repeated, provider reported misconfigured — which is the classification this shape of failure already had.

**Recorded as a paragraph in `docs/features/chat-generation.md` rather than as an ADR**, which the issue's acceptance criterion offers as the alternative. The reasoning: both surfaces are two paths under one address, reached with one credential over one transport under one resilience budget, and both are opened by the one existing client factory from the one already-pinned library. The provider set, the authentication shapes, and the boundary are unchanged, so nothing an ADR governs moved. Say the word if you read that differently and it becomes an ADR instead.

## Shape

Both clients are opened by `OpenAiCompatibleClientFactory` and both return the same provider-neutral `IChatClient`, so `ResilientChatClient`, `BudgetedChatClient`, and the composed agent are unchanged by the choice and never learn which surface they are speaking to. The API sits on `ChatEndpoint` rather than beside the generation parameters, because it is part of *where* a request goes.

One mapping — `ChatGenerationParameterMapping` — now builds the options both the single-request adapter and the answering run send. That removes the duplication that would otherwise let a parameter reach one path and not the other, so the effort rides every turn of the tool loop rather than only the first.

## One consequence, documented rather than hidden

The responses API reports an outcome instead of a finish reason: a response that stopped early says why, and one that simply finished names nothing. So over that surface a normal completion arrives as `Unreported` rather than `Completed`, which is the honest reading — claiming the model finished would state something the surface never said. A truncation and a content filter still arrive named, so neither is ever repeated as though it were a transport fault. Nothing acts on the stop reason beyond a log record.

## Breaking changes

**Configuration schema.** `Chat:ReasoningEffort` takes the provider's own level rather than an enum name. The key is new in this change, so nothing deployed carries a value yet; an operator following an early draft of it writes `xhigh` where they wrote `ExtraHigh`, and the other four are unchanged in effect once lower-cased.

Nothing else breaks: `Chat:Api` is additive and defaults to today's behaviour, and a deployment writing neither key sends the request it sent before. The MCP tool contract, the database schema, and the deployment contract are untouched. `ChatEndpoint` gains a member and `ChatGenerationPlan.Create` gains a parameter; both types are internal API with no external consumer.

## Verification

- `scripts/verify-full.sh` → exit 0. 4378 unit tests pass, 127 workflow assertions pass, aggregate line coverage 95.21% against the 85% gate.
- The wire is asserted rather than the intent: the effort's exact text on both surfaces, including a level no build knows; the absence of any reasoning member when nothing is declared; the request path per API; and the three outcomes the responses surface reports.
- A test runs the composed answering agent over a real provider client and a fake transport and asserts the effort on the wire of **every** turn. Nothing weaker can — the level is stated through a request hook rather than a value, so a substituted client only sees that the hook exists — and the turn after a retrieval is exactly the turn a provider refusing tools beside an unstated effort would reject.
- The shape check is anchored `\A`/`\z` rather than `^`/`$`, because `$` in .NET also matches before a trailing newline and a value provisioned from a file ends in one. `high\n` passed the first version of that check; both it and `high\r\n` are now in the suite.
- Both provider-contract tests run over **both** APIs, so a requested run makes four chat calls: two answers and two refusals. A refusal is answered before a model runs, so the second surface costs a round trip and no tokens. **Not run here** — the suite is manual and paid.

## One thing worth your eye

`THIRD_PARTY_LICENSES.md` carried a claim that stopped being true before this change: the OpenAI hosted-service row said *no caller composes one in this release*, which `ask_mail` made false. It is corrected here rather than left, because that row is the privacy record for the endpoint whose reachability this change alters, and the file was already being edited two rows above for the same reason.

Closes #622


tests/Host.UnitTests/Configuration/Chat/ChatModelOptionsTests.cs
```diff
@@ -280,6 +283,49 @@ public void Validate_UnsetSamplingParameters_AreAccepted()
         },
     };
 
+    /// <summary>
```
Krzysztof318, line 288: cover also responses api with integration tests, so both

src/Host/Configuration/Chat/ChatModelOptions.cs
```diff
@@ -118,6 +140,7 @@ public IEnumerable<ValidationResult> Validate(ValidationContext validationContex
             // wrote them out.
             if (this.Model.Trim().Length > 0
                 || this.Address.Trim().Length > 0
+                || this.ReasoningEffort is not null
                 || this.ApiKey is not null
                 || this.EntraCredential is not null
                 || this.RelevanceFilter.Enabled)
```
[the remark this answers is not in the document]
Krzysztof318: Fixed in 9a22487 — `Api` now joins the settings-without-an-alias guard as `this.Api != ChatProviderApi.ChatCompletions`, with a test case added to `Validate_SettingsWithNoAlias_AreRefusedRatherThanIgnored`.

You caught the gap in my own reasoning rather than an oversight, which is worth recording: I did consider `Api` for that guard and excluded it because it has a usable default, so a section that wrote `ChatCompletions` is indistinguishable from one that wrote nothing. That is true and it is not the test the guard applies — the guard asks whether the write was unambiguous intent, and a value *other than* the default is exactly that, because nobody writes `Responses` by accident. The comment above the condition said "every member checked here has no useful default", which was a rule that happened to fit the members present rather than the reason they were there; it now states what each member actually contributes, so the next member is judged by intent rather than by whether it resembles the existing ones.

docs/operations/configuration-reference.md
```diff
@@ -335,9 +335,11 @@ endpoint.
 | `Chat:Alias` | string | *(empty)* | writing one is what configures a chat provider at all; unique across every AI endpoint the deployment declares, embedding endpoints included. A section carrying other settings without it is refused rather than ignored | restart |
 | `Chat:Model` | string | — | required once an alias is written; what a request is routed to, which for a cloud deployment is the deployment's own name rather than the vendor's model identifier | restart |
 | `Chat:Address` | string | *(empty)* | absolute HTTPS; empty uses the provider library's default. A cloud resource's OpenAI-compatible address ends in `/openai/v1/` | restart |
+| `Chat:Api` | enum | `ChatCompletions` | `ChatCompletions` or `Responses`; which of the provider's two request APIs a call goes to under the declared address. Declared rather than derived, because the routed model name is the operator's own deployment name and nothing about it says which paths the server serves. State `Responses` for a reasoning model that refuses function tools beside a stated effort; a server that does not serve that path answers *request refused* | restart |
```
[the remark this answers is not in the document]
Krzysztof318: Fixed in 9a22487 — `docs/operations/configuration-reference.md` § `Chat` now carries a paragraph under the key table stating that `ask_mail` requires a model accepting function tools, that a reasoning model refusing them beside an unstated effort needs `Chat:Api = Responses` with `Chat:ReasoningEffort` written (including as `None`), and that such a mismatch is not caught at startup but surfaces as *request refused* on the first question. It links to chat-generation.md for the fuller reasoning rather than restating it.

The finding is right on the criterion and on the reader. I had put the statement in the two feature pages on the view that the reference table states constraints and the feature pages state consequences — but the two settings are unusual in that their whole purpose is a consequence, and an operator who opens the reference to find out what `Chat:Api` is for would have met two rows explaining what the values are and nothing saying why either exists.

tests/Host.UnitTests/Configuration/Chat/ChatModelOptionsTests.cs
Krzysztof318: Done in 9a22487 — both provider-contract tests are now `RetryTheory` over `ChatCompletions` and `Responses`, so a requested run makes four chat calls: two answers and two refusals.

Two consequences worth your eye, since they change what a run needs and what it costs.

`MAILFATHOM_CHAT_API` is gone. I had added it so a run could be pointed at one surface; once both are covered it would be a second mechanism deciding a question the tests now own, and a run that set it would silently narrow the coverage you asked for. `MAILFATHOM_CHAT_REASONING_EFFORT` stays and still applies to both.

The declared model now has to serve both surfaces. That holds for the first-party and cloud endpoints because a contract request carries no tools, and it is the toolless request that makes chat completions work for a reasoning model at all — the `400` this issue is about is specifically about tools *with* an effort. So the model that motivated the issue is a valid `MAILFATHOM_CHAT_MODEL` for this suite. A self-hosted OpenAI-compatible server that serves only `/chat/completions` would fail the two responses cases, which I think is the honest reading rather than a gap: it is the suite reporting that this deployment cannot reach that surface.

The refusal cases are close to free — a refused credential is answered before a model runs, so the second surface costs a round trip and no tokens.

Not run here, as the suite is manual and paid; `run-integration-tests.sh` with `MAILFATHOM_AI_CONTRACT_TESTS=true` is what proves it.

src/AI/Chat/ChatReasoningEffort.cs
```diff
@@ -0,0 +1,38 @@
...
+/// unstated effort names two ways out, and this is the one that keeps the tools. A run is a tool loop by construction —
+/// the model asks for mail, retrieval answers, the model writes — so removing the tools would remove the capability
+/// rather than work around the refusal.
+/// </para>
+/// <para>
+/// The members are what the neutral AI abstraction the request is built through can carry, rather than one provider's
+/// vocabulary. A value MailFathom accepted but could not send would be a setting an operator writes and nothing reads.
+/// </para>
+/// </remarks>
+public enum ChatReasoningEffort
```
Krzysztof318: don't hardcode values, new model can change this list

Krzysztof318: Done in 29db49d — the enum is gone and the effort is now the provider's own word, carried through as written.

You are right and the enum was the wrong shape for exactly the reason you give. Worse than I first saw, in fact: the level was going out through the provider-neutral reasoning member of the AI abstraction, whose set is fixed when *its* package is compiled, so `minimal` was already unreachable — a level OpenAI ships today. `xhigh` is the same story one release earlier. Making it a set here would have meant a MailFathom release every time a model gained a level, which is the opposite of what `Model`, `Address`, and the budget already do in this section.

What changed. The effort is now stated through the client library's own request options, whose level type is built from a string, rather than through the neutral member. The two APIs frame it differently — a top-level member on chat completions, a reasoning block on responses — so the hook is chosen per API. Everything else on the request is untouched: the abstraction fills the members it owns over the top, so bounds, sampling parameters, the instruction, and the tools all still go out. I verified that on the wire rather than by reading the library.

Startup checks the shape and never the vocabulary — one word, letters and digits with single hyphens or underscores between, at most 32 characters. A list here could only be the list that was true when it was written, so a level the model does not know is a *request refused* rather than something MailFathom pre-empts.

`ChatProviderApi` I deliberately left a closed enum. It is not the same kind of thing: it selects which client this build constructs, so a third value would be code rather than configuration, and a model cannot add one. Say the word if you want that opened too and I will, but I think a string there would be a setting that can only be wrong.

Two things worth flagging on top of the ask.

The shape check is anchored `\A`/`\z` rather than `^`/`$`. `$` in .NET also matches before a trailing newline, so `high\n` — which is what a value provisioned from a file looks like — passed the first version of the check and would have gone out on the wire. The test I wrote caught it before the gate did; both the `\n` and `\r\n` cases are in the suite now.

There is a new test that runs the composed answering agent over a real provider client and a fake transport and asserts the effort on the wire of **every** turn. Nothing weaker can: the level is now stated through a hook rather than a value, so a substituted client only sees that the hook exists. The turn that writes the answer is the one after a retrieval, which is precisely the turn a provider refusing tools beside an unstated effort would reject — a run that dropped the effort on turn two would fail in the one place this whole feature exists to fix.

This is a configuration break worth naming in the release notes: an operator who wrote `ExtraHigh` now writes `xhigh`. The other four are unchanged in effect once lower-cased.

PR #644, opened 2026-08-09
[#643] Refresh get-started-contributors, and let it be run again to update a machine

### LINKED ISSUE #643: Refresh get-started-contributors against what the repository now is, and let a configured machine be brought back into line
`get-started-contributors` was written on 2026-08-06 (#484) and last touched on 2026-08-08 (#617). The repository moved
underneath it in the days since — the MCP surface it describes as three tools now serves five, `tools/` appeared beside
the directories it tours, and several of the files it points a newcomer at were rewritten — so a first-time contributor
is oriented against a shape that no longer exists and has to unlearn part of it in their first review.

It also has no memory of ever having run. Setting a machine up is asked for once, but the rules it installs into that
machine do not hold still: the SDK pin moves, the permission list grows, the block written into `CLAUDE.local.md` is
kept identical in two files, and board access is a grant the maintainer makes whenever they decide to. Today the only
way to bring a clone back into line is to run the whole first-run walkthrough again — welcome, orientation, and every
install check — which is long enough that nobody does it, so a machine configured in one week keeps a configuration the
next week's rules have already superseded.

**As a first-time contributor**, I want the orientation to describe the product and the tree that exist today, so
nothing I am told on my first day has to be corrected in a review.

**As somebody who set a machine up weeks ago**, I want to invoke the same skill again and be told only what has moved
since — the pins, the permission list, the role file, the board grant — so refreshing a clone costs minutes rather than
a repeat of the walkthrough I have already read.

**As the maintainer**, I want the record of a completed setup to live where every worktree of one clone reads the same
one, so a re-run started in a linked worktree does not report a configured machine as unconfigured and redo the lot.

## Acceptance

- The orientation names the MCP surface the deployment actually serves — `list_accounts`, `list_emails`,
  `search_emails`, `get_email_content`, `ask_mail` — and says what makes the last one conditional on configuration.
- Every file, section, and directory the skill points at resolves against `main`, including the tree tour, which names
  what is there now.
- A completed run writes a local record holding the moment, the base commit it was performed against, the resolved
  role, the board-probe outcome, and what each tool check answered with. It lives where all worktrees of one clone
  share it and where no commit can reach it, and it is never a tracked file.
- Invoking the skill with that record present runs a refresh rather than a first run: it names what moved between the
  recorded commit and the current base among the paths the setup depends on, leaves an installed tool alone while its
  check still passes, probes the role and the board again because both are grants rather than facts, rewrites the local
  instruction file and the permission list only where they differ from what the skill now says, and writes the record
  back.
- A refresh with nothing moved says so and stops without installing or rewriting anything. Deleting the record is the
  documented way to ask for the full first run again.
- The welcome and the six-point orientation belong to a first run; a refresh offers them rather than performing them.
- `docs/operations/agent-workflow.md` and `CONTRIBUTING.md` describe the two modes where they already describe the
  skill, and the skill's return block reports which mode ran and where the record is.

Krzysztof318:
`get-started-contributors` was written on 2026-08-06 and last touched on 2026-08-08, and the repository moved
underneath it: the MCP surface it introduced as three tools now serves five, and `tools/` appeared beside the
directories it tours. It also had no memory of ever having run, so bringing a machine back into line with rules that
had moved meant repeating the whole first-run walkthrough — long enough that nobody does it.

## What changed

**The orientation says what is there now.** Five read-only tools — `list_accounts`, `list_emails`, `search_emails`,
`get_email_content`, `ask_mail` — with the reason `ask_mail` is advertised only where a chat model and an embedding
model are configured. The tree tour names `tools/` and says how Claude Code reaches the skills through the
`.claude/skills` symlink every clone carries.

**A completed run records itself**, in `mailfathom-setup.json` inside the clone's git directory, resolved through
`git rev-parse --git-common-dir`. That is what makes one clone hold one record however many linked worktrees read it —
the bare answer is relative in the checkout and absolute in a worktree, and both point at the same directory — and
nothing under `.git/` can reach a commit, so the record needs no ignore rule to stay out of one. It holds the moment,
the base commit, the role, the board-probe outcome, and what each tool check answered with; no token, no path outside
the clone, and a fact a run could not establish is `null` rather than a guess.

**A second invocation refreshes instead of repeating.** The welcome and the orientation are offered rather than
performed, and every step carries a **Refresh:** clause beside the existing **Owner's checkout:** ones. What moved
decides most of it — a diff of the recorded commit against the current base over the paths a local configuration is
written from — with three things that do not follow that diff: the role and the board probe run every time, because
neither is written in a commit, and a recorded commit the clone cannot reach is read as *everything may have moved*
rather than as nothing. Deleting the record is the whole of the reset.

**`jq` joins the toolchain table**, because the refresh reads that record, and takes the `THIRD_PARTY_LICENSES.md` row
it was already owed by `scripts/review-obligations.sh`, `scripts/test-agent-workflow.sh`, and eight workflows. MIT for
jq itself; the GitHub license API's `NOASSERTION` reflects the further permissive notices in its `COPYING` rather than
different terms.

No public surface changes: no tool contract, no configuration key, no database schema, and no deployment contract.
Nothing here edits a tracked file at runtime either — the skill still writes only outside the change, now including its
own record.

## Verification

`scripts/verify-full.sh` — green. 127 workflow contracts passed, 4563 unit tests passed, aggregate line coverage
95.14% against the 85% floor.

Closes #643


.agents/skills/get-started-contributors/SKILL.md
```diff
@@ -23,7 +24,56 @@ It is written for a contributor working through a fork, which is how every step
...
+repository-local tools move, the permission list grows, the block written into `CLAUDE.local.md` is kept identical here
+and in `CONTRIBUTING.md`, and access to the roadmap board is a grant the maintainer makes whenever they decide to. So
+this skill is invoked at least twice in a clone's life: once to configure it, and afterwards to bring that configuration
+back into line with what the repository now says. Repeating the whole first run to find out what moved is long enough
+that nobody does it, which is how a machine keeps a configuration the current rules have already superseded.
+
+A completed run therefore records itself, and whether that record exists is what picks the mode:
+
+```bash
+record="$(cd "$(git rev-parse --git-common-dir)" && pwd)/mailfathom-setup.json"
```
Krzysztof318, line 41: .gitignore new file?

Krzysztof318: No entry is needed, and that is the reason the record lives where it does. The path resolves inside the clone's own git directory, and git tracks nothing under it at all — a path there is not untracked-but-ignorable, it is outside the working tree, so `git add -A` cannot reach it and `.gitignore` has nothing to say about it. The alternative would have been a root file under the existing `*.local.json` rule, which works but relies on an ignore rule staying correct and puts one copy per worktree on disk; `--git-common-dir` gives one record per clone instead, which is what a linked worktree needs to read what the checkout wrote.

The sentence right after that code block states it — "Nothing under `.git/` can reach a commit at all, which is the other half of why the record lives there rather than beside `CLAUDE.local.md` under an ignore rule."

PR #673, opened 2026-08-10
[#549] Bump main version to 0.6.0

### LINKED ISSUE #549: Cut and publish the 0.5.0 release
## Context

**No numbered specification backs this issue.** It is the release itself: cutting, tagging, and publishing `0.5.0`
from the work placed in the `0.5.0` milestone. #526 was the same issue for the release before it.

The procedure is decided and recorded:
[ADR 0004](https://github.com/Krzysztof318/MailFathom/blob/main/docs/decisions/0004-versioning-and-release-policy.md)
settles what the number means and where it comes from, `docs/operations/release-procedure.md` records the sequence, and
`$prepare-release` prepares the two pull requests and prints the order they and the tag between them have to land in.

## What this release carries

Whatever is placed in this milestone. It is read from what merged since the previous tag at the moment the release is
cut, and stated in the `CHANGELOG.md` section the release pull request writes; this issue does not restate it.

## Acceptance

- The changelog pull request `[#<issue>] Prepare release 0.5.0` merges first; **its merge commit is what is tagged and
  published**, so the tagged tree carries the released changelog and the prose files naming the release they ship in.
- An annotated `v0.5.0` tag is pushed on that merge commit, by the owner. Nothing else pushes it.
- The multi-architecture image reaches **both** GHCR and Docker Hub under one digest, tagged `0.5.0`, with `latest`
  moved onto that digest in both; the packaged chart states `appVersion: 0.5.0`; the GitHub release carries
  `mailfathom-schema-0.5.0.sql` with its checksum and the `mfctl` binaries with theirs; the release notes are the
  changelog section rather than a list of pull-request titles.
- `SECURITY.md` names `0.5` as the supported line, and the line it replaces moves down a row.
- The version-bump pull request `[#<issue>] Bump main version to 0.6.0` merges after the tag and closes this issue.
- The published image and chart are installed and verified against a real deployment before the release is announced.

Krzysztof318:
Closes #549.

Part two of cutting `0.5.0`. **It merges after the tag, never before**, so `main` returns to naming the *next* release rather than the one just published — and it is what closes the tracking issue, because a release is finished when `main` names `0.6.0` rather than when the changelog merged.

## Merge order

1. #672 — `[#549] Prepare release 0.5.0`. Its merge commit is what gets tagged and published.
2. The annotated `v0.5.0` tag on that merge commit, pushed by the owner.
3. **This pull request.**

## What is in the diff

- `Directory.Build.props` — `<VersionPrefix>` from `0.5.0` to `0.6.0`. It is the only place a version is *written*: the image's tags and labels arrive as build arguments, and the chart's `version` and `appVersion` are both supplied at package time from the same declaration.
- Thirteen `packages.lock.json` files — every `MailFathom.*` project reference they record, from `[0.5.0, )` to `[0.6.0, )`. Regenerated with `dotnet restore MailFathom.slnx --force-evaluate`, which is where the number they hold stops being true.

`deploy/helm/mailfathom/Chart.yaml` is deliberately untouched: its `version` is a `0.0.0` placeholder and it declares no `appVersion`, both of which the release run supplies. The prose files naming the current release belong to #672 for the reason the ordering rests on — this pull request merges after the tag, so a `README.md` corrected here would have been wrong in the tagged tree.

## Verification

The transitive closure did not move. `git diff -U0 -- '**/packages.lock.json'` filtered of `MailFathom.` lines prints nothing, so this is a project-version bump and no dependency change is hiding inside it:

```
$ git diff -U0 -- '**/packages.lock.json' | grep -E '^[+-][^+-]' | grep -v 'MailFathom\.'
(no output)
```

All 45 project references name `0.6.0`, checked against the new version rather than the old one:

```
$ git grep -hoE '"MailFathom\.[A-Za-z]+": "\[[0-9]+\.[0-9]+\.[0-9]+' -- '**/packages.lock.json' | grep -v '0.6.0'
(no output)
```

`scripts/read-declared-version.sh` reports `0.6.0`, and `scripts/verify-full.sh` passed.

**Nothing in that gates the lock-file half**, and it is worth saying so rather than letting a green run imply it. Locked-mode restore does not compare a project reference's version, so lock files left naming `[0.5.0, )` would have restored green in both verification scripts and in every workflow — `NU1004` is what a moved *package* pin produces, not a stale project version. The two greps above are what actually verified this half, by inspection, which is why `docs/operations/release-procedure.md` step 3 keeps it a step rather than a check. What a skipped regeneration costs is paid later: the next `--force-evaluate` writes the truth, so the next pull request moving a central pin would arrive carrying this bump beside it, and its reviewer would have to tell the two apart.

`AppHost` and `IntegrationTests` carry no lock file and were not given one: the Aspire SDK picks part of their graph from the host platform's runtime identifier.

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -36,7 +36,7 @@
        same assembly identity. InformationalVersion is left to the SDK, which is what composes VersionPrefix,
        VersionSuffix, and SourceRevisionId into the single string the runtime reports. -->
   <PropertyGroup>
-    <VersionPrefix>0.5.0</VersionPrefix>
+    <VersionPrefix>0.6.0</VersionPrefix>
     <AssemblyVersion>$(VersionPrefix)</AssemblyVersion>
     <FileVersion>$(VersionPrefix)</FileVersion>
   </PropertyGroup>
--- a/src/AI/packages.lock.json
+++ b/src/AI/packages.lock.json
@@ -428,7 +428,7 @@
       "MailFathom.Application": {
         "type": "Project",
         "dependencies": {
-          "MailFathom.Domain": "[0.5.0, )"
+          "MailFathom.Domain": "[0.6.0, )"
         }
       },
       "MailFathom.Domain": {
--- a/src/Cli/packages.lock.json
+++ b/src/Cli/packages.lock.json
@@ -53,7 +53,7 @@
       "MailFathom.Common": {
         "type": "Project",
         "dependencies": {
-          "MailFathom.Domain": "[0.5.0, )"
+          "MailFathom.Domain": "[0.6.0, )"
         }
       },
       "MailFathom.Domain": {
--- a/src/Host/packages.lock.json
+++ b/src/Host/packages.lock.json
@@ -598,8 +598,8 @@
         "type": "Project",
         "dependencies": {
           "Azure.Identity": "[1.21.0, )",
-          "MailFathom.Application": "[0.5.0, )",
-          "MailFathom.Domain": "[0.5.0, )",
+          "MailFathom.Application": "[0.6.0, )",
+          "MailFathom.Domain": "[0.6.0, )",
           "Microsoft.Agents.AI": "[1.15.0, )",
           "Microsoft.Extensions.AI": "[10.8.1, )",
           "Microsoft.Extensions.AI.Abstractions": "[10.8.1, )",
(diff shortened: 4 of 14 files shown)


Krzysztof318:
Correct, and the claim was wrong rather than merely loose. `NU1004` is what a moved package pin produces when the lock file was not regenerated; a project reference records `"type": "Project"` with no version, and the `[x.y.z, )` lines the bump rewrites are transitive entries that locked-mode restore never compares. Lock files left naming `[0.5.0, )` would have restored green in both verification scripts and in every workflow, so the body claimed a gate that does not exist — and the merge commit message would have carried that claim.

Body corrected. It now states that `scripts/verify-full.sh` passed, that nothing in it gates this half, that the two greps are what verified the regeneration by inspection, and what a skipped regeneration actually costs: the next `--force-evaluate` writes the truth, so the next pull request moving a central pin arrives carrying this bump beside it and its reviewer has to tell the two apart.

No file changed, so the head stays `4e81a9f` and the checks that passed on it still describe the diff.

PR #698, opened 2026-08-11
[#600] Record which provider endpoints were checked, and what each check rests on

### LINKED ISSUE #600: Verify and declare the provider endpoints supported out of the box
## Context

"It speaks the OpenAI wire protocol" is a claim about somebody else's service, and it goes stale without anybody here touching anything. A provider can expose OpenAI-compatible chat completions and no embeddings route at all; can expose both and reject the `dimensions` parameter this configuration may send; can return vectors of a width its own documentation names differently; can require a header the SDK does not send. None of that is visible from the compatibility claim, and all of it decides whether an operator's configuration works.

So this issue produces a list with a verification behind each row, rather than a longer list of names. For each candidate: which roles it serves, what address it is reached at, which credential shape it takes, whether it honours a requested vector width, and — the part that costs the work — evidence it was actually exercised rather than read about.

The list also has to say what it is *not* promising. A service on it has been checked at a point in time by a project that does not control it; a service absent from it may well work. Those are two different statements from "supported" and "unsupported", and stating them as such is what keeps the list honest a year later.

Candidates worth checking, from the shape of the mechanism rather than from any commitment: vendor endpoints with a compatibility layer such as Google Gemini, Mistral, Cohere, and xAI; aggregators and hosts such as Hugging Face's inference router, Groq, Together, Fireworks, DeepInfra, Nebius, and OpenRouter; and the self-hosted servers #599 makes reachable — Ollama, llama.cpp, vLLM, LM Studio, LocalAI, and text-embeddings-inference. The embedding role is the one to check first in every case, because it is the one a compatibility layer most often leaves out.

## User stories

**As an operator choosing a provider** I want to know which endpoints this project has actually run against, **so that** I can tell a checked claim from a plausible one.

**As a maintainer** I want the list to carry what was verified and when, **so that** a report that something stopped working is answered against a record rather than against a memory.

## Acceptance

- Each listed endpoint carries the roles it serves, the address shape, the credential shape, whether it honours a requested vector width, and what was exercised to establish it.
- The embedding role is verified separately from the chat role for every candidate, because a compatibility layer commonly serves one and not the other.
- Anything that turned out not to work is recorded with what failed, so the next person does not re-derive it.
- The list states plainly that presence is a check at a point in time and absence is not a refusal.
- Each claim is verified against that provider's own current documentation, and the check is what the entry rests on rather than this issue's own text.
- A provider whose failure is a MailFathom bug rather than an incompatibility gets its own issue rather than a footnote here.

## Out of scope

Adding an adapter for anything that does not speak the OpenAI wire protocol. That is a decision before it is work, and #601 owns it.

Continuous verification. Nothing here runs on a schedule or in a pipeline; the suite that would need provider accounts and provider spend is a separate question.

## Depends on

#599, for the half of the list that is self-hosted. The rest can be verified before it lands.

Krzysztof318:
## What this changes

"It speaks the OpenAI wire protocol" is a claim about somebody else's service, and it goes stale without anybody here
touching anything. It also answers none of the questions that decide whether an operator's configuration works: whether
the service serves an embeddings route at all, at which address, against which credential, and whether it honours the
vector width the declaration asks for. The two feature pages said only that compatibility is unverified — true, and it
leaves the operator with nothing to act on.

`docs/operations/provider-endpoints.md` is the register that answers it, one row per service, with the evidence on the
row rather than in a footnote. Nineteen entries in three groups — vendor compatibility layers, aggregators and hosted
open models, and the servers an operator runs themselves — each carrying the roles it serves, the base path `Address`
writes, the credential shape, whether `SupportsRequestedDimension` may stay at its default, and whether the row rests
on a call through this project's own adapter or on the service's own documentation read on a stated date.

## What the check actually found

The embedding role was verified separately from the chat role for every candidate, because that is the half a
compatibility layer most often leaves out, and it turned out to be the half that most often fails:

- **No OpenAI-compatible embeddings route at all** — Groq, xAI, and the Hugging Face Inference Providers router. The
  last is the one most likely to be assumed wrong: the router serves embedding models, and its OpenAI-compatible
  surface does not, which its own documentation states outright.
- **The requested-width parameter is not honoured** — Cohere lists `dimensions` among the parameters its compatibility
  API does not support; Mistral accepts a requested width under the name `output_dimension`, which is not the member
  that is sent; Gemini and Together document an embeddings request of `model` and `input` and nothing else; Text
  Embeddings Inference publishes no width parameter on the route at all. `SupportsRequestedDimension` defaults to
  `true`, so every one of those needs it written `false`.
- **It fails loudly rather than quietly on vLLM**, which accepts `dimensions` and refuses the request for a model that
  was not trained for Matryoshka representation — so the value follows the model an entry names, not the server.
- **Fireworks honours it per model** and documents `normalize` as defaulting to false, which is a declaration to check
  against `NormalizeVectors` rather than to assume.

## What the page refuses to imply

Presence is a check at a point in time by a project that controls none of these services, and absence is not a refusal
— both stated on the page rather than assumed. Naming a service is not adopting one, which is why none of the
seventeen unadopted entries takes a row in `THIRD_PARTY_LICENSES.md`; the register's own rule excludes a component that
is "merely named somewhere". And declaring any of them sends mail text out of the deployment, which the page states as
the decision it exists to inform rather than one it makes.

The limit of the instrument is on the page rather than behind it: a contract run reads its key variable as required and
presents what it reads, so it exercises the `ApiKey` shape and never `Unauthenticated` — which is why every
self-hosted row reads as documented rather than called.

## Breaking changes

None. No configuration key, tool contract, database column, or deployment contract moves; nothing outside `docs/`
changes at all.

## Verification

- `scripts/verify-full.sh` — green, 4853 tests, aggregate line coverage 95.31%, workflow contracts run because the
  change reaches beyond C#.
- `scripts/build-docs-site.sh` — builds with the one pre-existing `CHANGELOG.md` warning and no new link failure.
- `scripts/review-obligations.sh` — the new page carries a `describes:` marker resolving through git, joins
  `docs/operations/toc.yml`, and is reached from `embedding-generation.md`, `chat-generation.md`, and
  `configuration-reference.md`. No source changed, so no test or register row is owed.

Closes #600

### DIFF
--- a/docs/features/chat-generation.md
+++ b/docs/features/chat-generation.md
@@ -69,7 +69,9 @@ OpenAI-compatible, so a cloud deployment is that same client pointed at the reso
 the deployment's name as the routed model. A third-party service speaking the same protocol is declared by the same
 values and reaches the same code — and "compatible" is that service's claim about itself rather than something checked
 here, which [embedding generation § Compatible is not
-verified](embedding-generation.md#compatible-is-not-verified) states in full for both roles.
+verified](embedding-generation.md#compatible-is-not-verified) states in full for both roles. [Provider
+endpoints](../operations/provider-endpoints.md) is where a service somebody checked is recorded, including which of the
+two request APIs below it was found to serve.
 
 There is no vendor and no model identity in the declaration beside the routed name, and that is the difference from an
 embedding endpoint rather than an omission. A vector is stored and later compared against other vectors, so which model
--- a/docs/features/embedding-generation.md
+++ b/docs/features/embedding-generation.md
@@ -202,6 +202,13 @@ name matches one elsewhere may answer at a different width or with vectors that 
 surfaces as a *request refused* or an *unexpected vector shape* on the first call rather than at startup, because
 nothing here can ask a service what it serves without paying for a request.
 
+What an operator has instead of a check at startup is a record of checks somebody made. [Provider
+endpoints](../operations/provider-endpoints.md) is that record: which roles each checked service serves, the address
+and credential its entry writes, whether it honours a requested width, and — on every row — whether the entry rests on
+a call through this adapter or on the service's own documentation read on a stated date. Presence on it is a check at a
+point in time and absence from it is not a refusal, which is the same statement this section makes and the reason the
+two are one page apart rather than one list.
+
 ## What an address has to be
 
 **Absolute, and HTTP or HTTPS.** Startup refuses anything else, naming the endpoint alias and the rule.
@@ -259,7 +266,9 @@ into this process or shipped in its image — that would make MailFathom respons
 hardware sizing, and a second failure domain inside its own process, for a capability an operator gets by starting one
 container beside it. Which servers the project has actually exercised is stated under [compatible is not
 verified](#compatible-is-not-verified), and that set is unchanged by this: the mechanism reaches a service, and reaching
-it is not a claim that it was tested.
+it is not a claim that it was tested. [Provider endpoints § a model server you run
+yourself](../operations/provider-endpoints.md#a-model-server-you-run-yourself) is the register of the servers somebody
+checked, with the address shape and the requested-width answer each was found to have.
 
 ## Authentication has three shapes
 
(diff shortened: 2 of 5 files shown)


Krzysztof318:
One finding of this verification pass is MailFathom's own rather than a provider's, so it went to its own issue rather than staying a footnote on the page: a contract run reads its key variable as required and therefore never exercises the `Unauthenticated` credential shape — the one every self-hosted server presents and the only one a contributor can try without an account. https://github.com/Krzysztof318/MailFathom/issues/699 owns closing that. The page states the limit as implemented behaviour and carries no intent to fix it, which is why the six self-hosted rows read as documented rather than called.

Issue #458, opened 2026-08-05
Give mfctl an on-demand rule run and an explainable run history


Krzysztof318:
## Context

Rules run without anyone watching, and they change mail on the server. That makes two things necessary that ordinary features can leave implicit: a way to make them run over mail that is already there, and a way to find out afterwards what they did.

**`mfctl` runs rules; it does not author them.** A rule lives in configuration, so creating, editing, enabling, and disabling one is an edit to a file and a reload, and there is deliberately no command for any of it. What the operator needs from the CLI is the trigger #455 built — apply the rules to this account's mail now, over everything stored rather than only over what arrives next — and the ability to read back what happened. Listing the loaded rules is part of that reading: it answers *did my reload take effect and is my file valid*, which is otherwise something an operator infers from the absence of a log line.

**The history is a record of decisions, not a second copy of the mailbox.** Draft section 21.5 says what explainability has to cover — which rule version matched, which deterministic facts were used, which action was attempted, and why it succeeded, failed, or is waiting — and the clause about not storing unnecessary message content is the design constraint rather than a caveat. What it records is identity and outcome: message identifiers, the rule and the configuration revision it came from, the condition's result, the actions and their results, timing, and the execution identity.

A condition is one NCalc expression rather than a list of predicates, which sharpens that into a rule. There is no set of individually named conditions to report as having matched; what there is, is the expression's outcome and the facts the evaluation resolved along the way. **Those facts are recorded by name and never by value.** A resolved fact is a sender address, a subject, a header, or a span of extracted text — recording what it evaluated to is exactly the second copy of the mailbox this history exists not to be. The configuration revision is what makes the names sufficient: the expression is retrievable from the revision the run was bound to, so `subject`, `senderDomain`, `hasAttachment` and a true result reconstruct the reasoning without reconstructing the mail.

**It does not restate what the mutation trail already holds.** What became of a change on the server — attempted, converged, failed, given up on — is already recorded against the mutation, with its own audit trail and its own retention. The rule history records the decision and the request it produced, and points at that record rather than copying its lifecycle into a second place.

The commands go through the administrative endpoint that `mfctl` already uses, with its existing authentication.

## User stories

**As a mailbox owner** I want to apply my rules to a whole mailbox from the command line after I edit them, **so that** a rule I just wrote reaches the mail I already have.

**As a mailbox owner** I want to see why a message was filed, copied, deleted, or marked read, **so that** automation over my mailbox stays something I can audit.

[...]

PR #727, opened 2026-08-12
[#708] Declare per rule which triggers may run it, with an empty set meaning a requested run only

### LINKED ISSUE #708: Declare per rule which triggers may run it, with an empty set meaning mfctl only
## Context

#455 builds the two ways a rule pass starts: a message committed by synchronization, and an on-demand run over a whole account. Every rule is reached by both, and nothing a rule declares says otherwise. That is the gap this issue closes.

**A rule declares which triggers may run it.** `Triggers` is a list on the rule, beside `Accounts` and `Enabled`, and it holds the automatic triggers that rule participates in. A rule takes part in the occasions it names and in no others, so an absent key and `[]` say the same thing: a rule nothing fires by itself, which a whole-mailbox run applies.

> **Correction, decided during implementation:** this originally read that an absent key resolves to `[ "Arrival" ]` so that no rule written before the change would alter its meaning. There is no such rule — the feature has not shipped, nothing reads the key yet, and no deployment's configuration depends on the reading — so the default was reversed to nothing rather than kept for a compatibility that has nobody to protect. A rule meant to run over arriving mail declares `[ "Arrival" ]`. This is a break of the configuration schema and is recorded as one in the pull request, with the operator's action.

**An empty list is the point of the shape.** `"Triggers": []` is a rule that never fires by itself: it stays in the bound set, it is validated like every other rule, and the only thing that runs it is the on-demand run an operator starts from `mfctl`. That is what a periodic housekeeping rule wants — file everything older than a quarter, delete what a mailing list left behind — where firing on each arriving message is either useless or the thing the owner is afraid of. Today the alternative is `Enabled: false` and an edit-reload-edit-reload cycle around every run, which is a configuration change to perform an operation.

**The on-demand run is never a member of the list.** An operator asking for a run is the request itself, so a rule declining to run because it had not opted into being asked would be surprising in the one place surprise is least affordable. The list therefore governs automatic firing only, and the on-demand run applies the whole set, manual-only rules included. Selecting a subset of rules for one run is a different feature and is not proposed here.

**The list is a list rather than a single value**, for two reasons that hold now: an empty set is exactly what expresses *manual only*, where a single-valued key would need a magic `None` beside the real values, and the trigger set is open at the top — #455 records that a scheduled periodic scan was refused *for now* and that a third trigger would be an issue rather than a redesign. A rule that then wants both arrival and a schedule already has the shape to say so.

**`Arrival` rather than `Push`.** `Push` in this configuration already names an account's synchronization mode, and a rule is not affected by whether its account is watched or polled — it runs when synchronization commits a message either way. Naming the trigger after the moment rather than after the transport keeps one word from meaning two things in one file.

**Distinct from `Enabled`, and the documentation has to say so.** `Enabled: false` leaves the rule out of the bound set entirely, so it costs nothing and cannot run at all. `Triggers: []` keeps it in the set, keeps it validated, and keeps it runnable on request. Two adjacent keys that both sound like off, with different meanings, is precisely where a feature page earns its place.

This is not folded into #455 because it is a declaration-surface change: a key on the rule schema, its own validation, its own contribution to the rule set's revision identity, and its own page material. #455 is claimed and builds the triggers themselves; this filters them.

## User stories

**As a mailbox owner** I want a rule that only ever runs when I ask for it, **so that** quarterly housekeeping over my whole mailbox is something I trigger rather than something that fires on every message that arrives.

**As a mailbox owner** I want the rules I already wrote to keep running on arriving mail without editing them, **so that** the new key is an addition to the schema rather than a migration of my configuration.

**As the operator of my own instance** I want a mistyped trigger name to be refused when the configuration is read, **so that** a typo cannot silently turn an automatic rule into one that never fires.

## Acceptance

- A rule declares `Triggers`, a list of trigger names, and the value set is closed with exactly one member today: `Arrival`, meaning a message committed by the account's synchronization run.
- A rule that does not declare the key takes part in no automatic occasion, which is what `[]` says as well; a rule meant to run over arriving mail declares `[ "Arrival" ]`.
- `"Triggers": []` is accepted and means the rule never runs automatically. It is bound, validated, and reported like any other rule, and the on-demand run from `mfctl` runs it.
- The on-demand run applies the whole rule set for the account, manual-only rules included, and no run selects rules by name.
- An unrecognized trigger name is refused when the configuration binds, naming the rule and the value. It is never dropped, and a list whose only entry was mistyped never arrives as an empty list — silently turning an automatic rule into a manual one is the failure this check exists for.
- A duplicate entry in the list is refused with the same message shape, because the value is a set.
- `Triggers` is part of what the rule set's revision identity is derived from, so changing it changes the revision exactly as changing a condition does, and a run in flight stays bound to the revision it started under.
- A rule the current trigger does not reach is not evaluated and records no outcome, the same way a rule scoped to another account does not — the history stays a record of decisions rather than of rules that were passed over.
- Unit tests cover the absent-key default, the empty list on both triggers, the refusal of an unknown and of a duplicated value, the revision change, and that a manual-only rule runs in an on-demand pass and not in an arrival pass.
- The feature page documents the key, its default, the empty list, and the difference between `Triggers: []` and `Enabled: false`, with a worked rule that is manual only. The configuration reference lists the key with its type, default, and constraint, and the administration page states that an on-demand run is how a manual-only rule is run.

## Out of scope

Any new trigger, including a scheduled one — #455 records why none exists yet. Running a named subset of rules in one on-demand pass. Any trigger for a rule that is not one of the two #455 builds.

## Depends on

#455 — the triggers this filters.
#458 — the on-demand run that is the only way a manual-only rule runs.

Krzysztof318:
Closes #708

A rule now declares `Triggers`, the automatic occasions it takes part in. `Arrival` is the only member today, and a
rule takes part in the occasions it names and in no others: an absent key and an empty list both say that nothing
fires the rule by itself, which leaves it bound, validated, and applied by a whole-mailbox run rather than by arriving
mail.

## What it does

- **`MailRuleTrigger`** is a closed enumeration: the name is what an operator writes in configuration and what the rule
  set's revision is derived from, so it must survive a rename of the C# member.
- **`MailRuleReach`** tells a triggered walk from a requested one. A whole-mailbox run is deliberately not a trigger — an
  operator asking for a run is the request itself — so it reaches every rule of the set, manual-only rules included, and
  nothing selects a subset of rules for one run.
- **A rule a walk's reach does not hold is passed over and records no outcome**, exactly as a rule scoped to another
  account is, and it cannot end the pass whatever `StopWhenMatched` says.
- **The body-text question is asked per walk** rather than per pass, so a manual-only rule naming `bodyText` does not
  hold arriving mail behind extraction.
- **`Triggers` is part of the revision.** Withdrawing a rule from arrival moves it; the case a trigger's name was
  written in does not, because the digest reads the trigger rather than the text naming it.
- **The reporting surface says what runs a rule.** `MailRuleResponse` carries the triggers, and `mfctl rules list` and
  `mfctl rules show` print a `Runs on:` line. A rule nothing fires by itself says what does run it instead of reporting
  an empty list, because it is otherwise indistinguishable from a rule that never matched.

## The trap the shape exists for

**The elements are text, not the trigger.** The binder drops an element it cannot convert, so a mistyped name in a
bound list of triggers would arrive as a shorter list — and a list whose only entry was mistyped would arrive as the
empty one, silently turning an automatic rule into a manual one. Read as text, every name reaches validation, an
unknown or repeated one is refused naming the rule and the value, and the mapper refuses a set that reaches it with
one. `MailRuleOptionsBindingTests` is where that claim is checked rather than assumed.

## Contract notes

- **Breaking, configuration schema.** `MailRules:Rules:<n>:Triggers` has no default: a rule that should run over
  arriving mail declares `[ "Arrival" ]`. **Operator action:** add the key to every rule meant to fire on arrival. A
  rule left without it stays in the set and is applied by `mfctl rules run` alone; `mfctl rules list` shows which is
  which on its `Runs on:` line. Nothing has shipped that reads the key, so this affects a configuration written against
  this branch rather than against a release.
- **Every rule set's revision moves**, because the digest gains a field. A whole-mailbox run outstanding across the
  upgrade therefore ends as `superseded` and is asked for again — the documented behaviour for a rule set that moved
  while a run was in flight.
- `MailRule.Create` and `MailRuleDeclaration` gain a parameter, and `MailRuleSetEvaluator.EvaluateAsync` gains the
  reach. Both are internal cross-project contracts rather than published surfaces.

## Verification

- `scripts/verify-full.sh` — green on the rebased branch (5525 unit tests, 148 agent-workflow assertions, formatting
  made no change).
- `dotnet msbuild .config/CodeCoverage.proj -t:Collect` — 95.31% aggregate line coverage, above the 85% gate.
- `$check-docs-licenses` — Docs pass, Changelog n/a, Licenses n/a (no new dependency, tool, service, image, or copied
  code).

## The acceptance item #458 owed

#458 has merged, so this change satisfies it here rather than carrying it forward: `docs/users/administering.md` and
`docs/operations/admin-endpoint.md` both state that a whole-mailbox run is how a rule naming no trigger is run.

https://github.com/Krzysztof318/MailFathom/issues/708

### DIFF
--- a/docs/features/mail-rules.md
+++ b/docs/features/mail-rules.md
@@ -23,6 +23,7 @@ MCP, not the administrative endpoint. An owner who wants to change what their in
         "Name": "supplier-invoices",
         "Accounts": [ "work" ],
         "Condition": "senderDomain == 'supplier.test' and attachmentCount > 0",
+        "Triggers": [ "Arrival" ],
         "Actions": { "MoveTo": "invoices", "MarkAsRead": true },
         "StopWhenMatched": true
       },
@@ -43,8 +44,10 @@ part that can be promised to carry no such thing when it reaches a log line.
 `Enabled` defaults to `true`. A rule switched off is left out of the bound set entirely, so it costs nothing and
 changes the set's revision exactly as deleting it would.
 
-`Triggers` defaults to `[ "Arrival" ]` and is what decides when a rule runs; [the section below](#which-triggers-run-a-rule)
-states what each way of writing it means, and why it is a different statement from `Enabled`.
+`Triggers` is what decides when a rule runs, and it defaults to nothing: a rule that should run over arriving mail says
+`[ "Arrival" ]`, and a rule that says nothing is one a whole-mailbox run applies. [The section
+below](#which-triggers-run-a-rule) states what each way of writing it means, and why it is a different statement from
+`Enabled`.
 
 ## Which accounts a rule applies to
 
@@ -77,23 +80,24 @@ decides *when* a rule is reached rather than which mail it matches once it is.
 
 | Written | What the rule takes part in |
 | --- | --- |
-| the key left out | `Arrival`, which is what every rule written before the key existed already did |
-| `[ "Arrival" ]` | The same, said explicitly |
+| `[ "Arrival" ]` | Every message the account's synchronization run commits |
 | `[]` | No automatic occasion at all: nothing fires the rule, and a whole-mailbox run is what applies it |
+| the key left out | The same as `[]` — a rule takes part in the occasions it names and in no others |
 
 **`Arrival` is a message the account's synchronization run has just committed.** It is named after the moment rather
 than after the transport, because `Push` already names a folder's
 [synchronization mode](imap-synchronization.md#choosing-the-mode-per-folder) and a rule is unaffected by which one its
 account uses: mail a polled run commits reaches this trigger exactly as mail a watched one commits does.
 
-**Leaving the key out changes no rule's behaviour**, which is what makes the key an addition to the schema rather than
-something to edit a configuration for. A rule that writes `[ "Arrival" ]` and a rule that says nothing are the same rule
-set, down to [the revision](#the-revision-a-pass-runs-under) both are identified by.
+**A rule that names no trigger runs on no arriving message.** Leaving the key out is a statement rather than an
+omission, and it is the same statement as writing `[]`: the rule is bound, validated, and reported, and a whole-mailbox
+run is what applies it. So a rule meant to file mail as it arrives writes `Arrival` — there is no occasion a rule joins
+without naming it.
 
-**A rule that declares `[]` is a rule you run rather than one that runs.** That is what periodic housekeeping wants —
-file everything older than a quarter, delete what a mailing list left behind — where firing on each arriving message is
-either useless or exactly what the owner is afraid of. The rule is bound, validated, and reported like any other, and
-[a whole-mailbox run](#running-the-rules-over-mail-you-already-have) is what applies it.
+**Such a rule is one you run rather than one that runs**, which is what periodic housekeeping wants — file everything
+older than a quarter, delete what a mailing list left behind — where firing on each arriving message is either useless
+or exactly what the owner is afraid of. [A whole-mailbox
+run](#running-the-rules-over-mail-you-already-have) is what applies it.
 
 **A rule a trigger does not reach is not evaluated and records no outcome**, exactly as a rule scoped to another
 account is not: it did not decline to match, it was not one of that pass's rules. It follows that such a rule cannot end
@@ -457,7 +461,8 @@ before it has already stored, so a provider redelivering a message or a synchron
 processing boundary than a clean run. And nothing an MCP tool does waits on a rule: reads are served from what is
 already stored, and a pass neither blocks one nor is blocked by one.
 
-**A rule applies to mail that arrives after the rule exists.** Each message is evaluated once, and the record of that
+**A rule declaring `Arrival` applies to mail that arrives after the rule exists.** Each message is evaluated once, and
+the record of that
 evaluation is what takes it out of the queue the next pass reads — so editing a rule changes what happens to the mail
 that arrives from then on and does nothing to the mail already in the mailbox. Mail an instance stored before it had
 rule evaluation at all is recorded as evaluated when the schema is applied, for the same reason: an upgrade must not
@@ -470,9 +475,10 @@ evaluated drains over as many runs as its size needs instead of holding up the r
 commits its evaluations together with the position they account for, so a restart resumes at the message nobody read
 rather than replaying a batch or stepping over one.
 
-**A message whose body text has not been extracted yet is skipped and stays eligible.** It applies only where the
-rules that walk actually runs for the account name `bodyText` — so a manual-only rule naming it holds nothing up on
-arrival, and a rule that names it on arrival does: such a message is left in the queue and evaluated once its text has been
+**A message whose body text has not been extracted yet is skipped and stays eligible.** It applies only where the rules
+that walk actually runs for the account name `bodyText`, which is decided per walk — so a rule only a whole-mailbox run
+applies holds nothing up on arrival, and a rule declaring `Arrival` does: such a message is left in the queue and
+evaluated once its text has been
 derived, rather than evaluated against a fact that would answer absent and then never reconsidered. Mail whose payload
 local storage has not had headroom for is waited on the same way, because a later run fetches it as soon as the ceiling
 permits. A message whose content will never yield text — one above the size limit, which every later run refuses for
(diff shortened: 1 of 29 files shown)


Krzysztof318:
Answering the P3 finding raised on both passes, which has no line to sit on.

The acceptance item is now recorded where it survives this issue closing, rather than only in the pull-request body:

- #708 carries a comment naming the item, why it is not delivered here, and what was delivered instead: https://github.com/Krzysztof318/MailFathom/issues/708#issuecomment-5267533357
- #458 carries a comment stating the sentence its administration page owes, so the item is checked against the issue that can actually satisfy it: https://github.com/Krzysztof318/MailFathom/issues/458#issuecomment-5267535589

No follow-up issue was opened, because #458 already owns both the `mfctl` command and the administration page the item belongs on, and a third issue would only duplicate what that one has to do.

Nothing in the diff changes as a result: `docs/users/administering.md` would have to describe a command that does not exist, which `docs/AGENTS.md` refuses, and the behaviour itself is documented on the feature page.

Krzysztof318:
Rebased onto `main` now that #458 has merged, and the acceptance item this pull request was carrying is delivered here rather than deferred.

- The walk's two vocabularies are collapsed: `MailRuleEvaluationPass` derives the `MailRuleReach` from #458's `MailRuleExecutionTrigger`, so `Arrival` runs the rules that declare it and a requested run applies the whole set.
- #458's reporting surface reports triggers as well: `MailRuleResponse` carries them, and `mfctl rules list` and `mfctl rules show` print a `Runs on:` line. A rule declaring `"Triggers": []` says what does run it instead of reporting an empty list, because such a rule is otherwise indistinguishable from one that never matched.
- `docs/users/administering.md` and `docs/operations/admin-endpoint.md` state that a whole-mailbox run is how a rule declaring `"Triggers": []` is run — the item recorded on #708 as owed to the administration page.

Verification: `scripts/verify-full.sh` green (148 workflow checks, 5527 tests, 0 failed), coverage 95.24% against the 85% floor.

docs/features/mail-rules.md
```diff
@@ -66,6 +70,58 @@ that applies to several.
...
 type, default, and constraint.
 
+## Which triggers run a rule
+
+`Triggers` is the list of automatic occasions a rule takes part in. It governs automatic firing and nothing else, so it
+decides *when* a rule is reached rather than which mail it matches once it is.
+
+| Written | What the rule takes part in |
+| --- | --- |
+| the key left out | `Arrival`, which is what every rule written before the key existed already did |
```
Krzysztof318: there is no rules without this key as the feature wasn't shipped yet. SO default empty, adn has to be specified manually to Arrival

Krzysztof318: Done — the key now defaults to nothing, and a rule takes part in the occasions it names and in no others. `[]` and an absent key say the same thing, so a rule meant to run over arriving mail writes `"Triggers": [ "Arrival" ]`.

What moved with it:
- `MailRuleTrigger.WhenNoneDeclared` is gone, and `MailRule.Create` resolves an absent list to the empty set.
- `MailRuleOptions.Triggers` is `IList<string>` defaulting to `[]`, the same shape `Accounts` uses. The array type existed only to tell a written `[]` from an absent key, and nothing rests on that distinction now; the elements stay text so an unreadable name reaches validation instead of being dropped.
- Validation is unchanged: an unknown or repeated name is still refused, naming the rule and the value.
- Documentation: the `Triggers` table, the anatomy paragraph, the worked rules (each one meant for arriving mail now declares it), the configuration reference default, and the admin/mfctl pages.
- Tests: the arrival-walk suites build their rules through an `ArrivalRule` helper that declares the trigger, and the tests asserting the old default now assert the new one.

Recorded as a breaking configuration change in the commit and in the pull-request body.

Krzysztof318:
P2 answered — issue #708 now records the reversal rather than contradicting it.

Its intro paragraph and the acceptance item both said an absent `Triggers` key resolves to `[ "Arrival" ]` so that no rule written before the change alters its meaning. Both now state the shipped reading — a rule takes part in the occasions it names and in no others — and the intro carries a short correction block naming why the default was reversed (nothing has shipped that reads the key, so the compatibility the old default protected has nobody to protect), that a rule meant to run over arriving mail declares `[ "Arrival" ]`, and that this is a configuration-schema break recorded with the operator's action in this pull request.

That is a fix outside the diff, so it cannot re-trigger the review: the head commit is unchanged at 68e7510f, and the finding had no line to sit on.

- https://github.com/Krzysztof318/MailFathom/issues/708

Issue #786, opened 2026-08-12
Point an operator and a user at the documentation for the version they are running


Krzysztof318:
## User stories

- As somebody holding only an image reference, I want `docker inspect` to name where that image's documentation is, so I do not have to work out which version's pages apply to it.
- As somebody who has just run `helm install`, I want the notes to hand me an address I can paste into an agent, rather than a repository path I have no checkout for.
- As somebody whose chat client has connected over MCP, I want the connection itself to say where the documentation is, so my agent finds it without me configuring anything.

## Why four surfaces rather than one skill

A skill reaches one harness and names whichever version was written into it. Each surface below already knows a version at the moment it speaks, and none of them asks the reader to install anything: the label and the chart notes reach whoever installs, `mfctl` reaches whoever administers, and the MCP session reaches whoever only ever uses the tools.

## The address every surface prints

One convention, shared with #785 and settled before either is worked: **the documentation directory of the version, `https://krzysztof318.github.io/MailFathom/<version>/`**, rather than a file inside it. The site already publishes one directory per documented version, so that address resolves today and this issue does not wait on anything. #785 then puts its map inside that directory and links it from the page, which is what carries an agent from this address to the artifact written for it — without a second change here.

## Acceptance

- `deploy/docker/Dockerfile` declares `org.opencontainers.image.documentation`, built from the `IMAGE_VERSION` argument beside the labels already there, and `publish-container-image.yml` asserts it off the published manifest as it asserts the others.
- The chart's `NOTES.txt` prints the documentation address for the version it installed. Its existing pointer to `docs/operations/database-schema.md` becomes an address too: a repository-relative path is unreachable for somebody reading `helm install` output, which is a defect on its own terms.
- `mfctl status` reports the address for the version **the deployment** reports rather than the one `mfctl` was stamped with, and says nothing about it where that version could not be read — the case `DeploymentVersionAgreement` already treats as absence of evidence rather than evidence of a break.
- The MCP server sets `ServerInstructions`, which the pinned SDK carries, so an initializing client learns the address for the running version. It carries one sentence and an address and no documentation content: this issue makes documentation findable, never serves it over the protocol.
- The address is derived from the version. No configuration key is added for it and no part of it is operator-editable.
- Unit tests cover the address a given version produces and the unreadable-version case.
- The page describing each surface states its new output.

## Related

- Shares the address convention above with #785 and nothing else, so the two are worked in parallel and merge in either order.
[...]

PR #793, opened 2026-08-12
[#785] Publish the documentation as artifacts an AI agent can read

### LINKED ISSUE #785: Publish the documentation as artifacts an AI agent can read
## User stories

- As somebody installing or operating MailFathom with an AI agent beside them, I want the agent to load a map of the documentation in full and then fetch only the pages that answer my question, so its answer comes from the page that owns the contract rather than from a search over fragments.
- As somebody using the MCP tools from a chat client, I want my agent to take the whole user path in one fetch, so asking what a tool bounds does not start with me finding pages for it.
- As the owner, I want a published page that no artifact lists to fail the build, so this cannot rot the way a hand-maintained index would.

## What is published

The site already holds one directory per documented version, so these are per-version artifacts and inherit that. `docs/` is roughly 1.76 MB of Markdown, which is why the map and the bundles are separate answers: the map covers everything and stays small, and a bundle covers one reader's path in full.

- **`llms.txt`** — the map. Every published page with its title and one line saying what it answers, ordered as the section's `toc.yml` orders it. Published at each version's root, and at the site root for the release the site opens on.
- **The Markdown source beside each page**, at the same path with a `.md` extension, so a link from the map fetches text rather than a rendered template.
- **Two bundles of the user guide**, because it serves two readers: the operator path — choosing an installation, getting started, mailbox providers, administering — and the mailbox-user path — connecting a client, using the tools. Together they are about 119 KB; separately each is what one reader needs and no more.

No bundle of the whole of `docs/`. At 1.76 MB nothing loads it, so it would be an artifact to keep correct that no reader ever benefits from.

## Acceptance

- The artifacts above are produced by the site build and published under every documented version, `latest` included.
- `scripts/test-agent-workflow.sh` fails a published page the map does not list and a map entry naming no page, the way it already fails both halves for `toc.yml`.
- Links inside the artifacts resolve under the rule the site already applies to a page's links.
- **The version's own landing page links the map.** That page is the address #786 prints from the image label, the chart notes, `mfctl`, and the MCP session, so it is what carries an agent arriving there to the artifact written for it. Without this the two halves never meet.
- `docs/operations/documentation-site.md` states what the build emits and what it deliberately leaves out; `docs/users/README.md` names the artifacts for a reader who wants to hand one to their agent.
- Any new script or generated asset carries the licensing header in the form its own readers parse.

## Related

- #786 is worked in parallel with this and merges in either order. The two share one decision, recorded on the parent: every surface there prints the version's documentation directory, which the site already publishes, and this issue links the map from that page.
- #787 waits on this one, because a mirror submitted earlier would index a shape about to change.

Krzysztof318:
Closes #785

The documentation site now publishes the documentation twice: as pages a person browses, and as the artifacts an AI agent reads. `docs/` is roughly 1.8 MB of Markdown, so the split is what a context window costs — the map covers everything and stays small, and a bundle covers one reader's path in full.

## What every built version now carries

| Artifact | What it is |
| --- | --- |
| `llms.txt` at the version's root | The map: every published page, its title, and one line saying what it answers, in the order the navigation puts it, following the [llms.txt convention](https://llmstxt.org/) |
| `<page>.md` beside each rendered page | The page's Markdown source at the same address with a `.md` extension, so a link from the map fetches text rather than a template |
| `llms-operator.txt`, `llms-mailbox-user.txt` | The user guide's two reading paths, each page of one path in one file — 78 KB and 41 KB against 11 KB for the map |

`scripts/compose-docs-site.sh` copies the default version's map to the site root, with every link rebased into that version's directory. A release built before these artifacts existed carries no map, and the site opens on the newest release — so the root map arrives with the first release that carries one, and the composition says so rather than failing a publish over a commit that could not have written it.

No bundle of the whole of `docs/`, and no Markdown for the generated API reference: both would be artifacts to keep correct that no reader benefits from.

## The map is the navigation

A page's title and its place in the map are the `name:` and the position it already has in a `toc.yml`; the line saying what it answers is a new `description:` beside them. That is one file to write when a page is added rather than two, and it is what stops the map from rotting — there is no second list to forget a page in. Every entry naming a page or a section now carries one; a group heading names no page and carries none.

Both halves of that agreement are gated, in the build and in the contract suite:

- a published page the map would miss,
- a map entry the version carries no page for,
- a table-of-contents entry with no `description:`.

`scripts/test-agent-workflow.sh` asserts the first two against this repository's own navigation and drives all three from fixtures, beside a contract that a bundled page's relative links are resolved for the move to the version's root.

## Links

`scripts/rebase-markdown-links.sh` is the one operation two callers share: a bundle sits at the version's root rather than in `users/`, and the root map sits above the version directory. A relative link is resolved for that move — `../operations/mcp-endpoint.md` in a user-guide page becomes `operations/mcp-endpoint.md` — and every absolute URL, scheme, root-relative path, and bare fragment is left exactly as written. Verified on the built site: every relative link in all three artifacts resolves against the version directory.

The version's landing page links the map, which is what makes the artifacts reachable from the address every surface prints (#786). docfx cannot resolve that link, because the artifacts are written into its own output after it finishes, so `scripts/build-docs-site.sh` exempts exactly those targets from the link gate — the artifact script's own check that every entry names a page the version carries is what covers them instead.

## Surfaces

No breaking change to any of the four public surfaces. The documented site gains files; nothing it already served changes address or content.

## Verification

- `scripts/verify-full.sh` — pass on the staged tree: 6079 unit tests, 162 workflow contracts, 0 failed.
- `scripts/build-docs-site.sh artifacts/docs-site` — pass end to end, with the artifacts written into the built version and the link gate green.
- `scripts/compose-docs-site.sh` against a composed probe site — the root map written and rebased.
- Every relative link in `llms.txt` and both bundles checked against the built version: none missing.

Issue: https://github.com/Krzysztof318/MailFathom/issues/785


Krzysztof318:
On the one thing the review could not verify — the shape of `agent_artifact_link_exemption` against docfx's actual diagnostic — here is the evidence it was missing.

The regex was written against a real docfx log rather than an assumed format, and its first version was wrong for exactly the reason the review suspected. docfx 2.78.5 writes one JSON object per line, and the diagnostic is:

```json
{"severity":"warning","message":"Invalid file link:(~/../docs/llms.txt).","file":"../docs/index.md","line":"37","code":"InvalidFileLink"}
```

The path in `message` is relative to the docfx configuration's own directory, so `docs/index.md` linking `llms.txt` produces `~/../docs/llms.txt` rather than `~/llms.txt`. A first attempt anchored on `~/llms` and `scripts/build-docs-site.sh` failed the build with the three diagnostics above; the committed pattern matches the file name at the end of the path instead, and the same build then passed. `Build latest` on this pull request is that build running in the pipeline, and it logs `Wrote llms.txt, 52 Markdown source(s), and 2 bundle(s) into .../site/latest`.

The exemption is deliberately keyed on the target rather than on the linking page — the sibling `changelog_link_exemption` matches `"file"` because there the *page* is what may not be corrected, whereas here it is the *target* docfx cannot see. A link to anything else still fails from any page, and `scripts/write-docs-agent-artifacts.sh` covers the exempted targets itself: it refuses a map entry the built version carries no file for, which includes the map and both bundles.

PR #860, opened 2026-08-15
[#859] Select the reviewer's Claude credential by profile

### LINKED ISSUE #859: Authenticate the reviewer with the Claude credential the selected profile owns
`Fathom review` authenticates with one repository secret, `CLAUDE_CODE_OAUTH_TOKEN`, named
literally in the three places the workflow touches a Claude credential: the action input that
spends it, and the two leak checks that refuse to publish anything containing it. The owner now
runs Claude Code under more than one profile, and the second profile's subscription has its own
token. Today reaching it means overwriting the one secret, which is a change every run inherits
and nothing records.

## User stories

- As the repository owner, I want to point the reviewer at my secondary Claude subscription by
  setting one repository variable, so that switching which subscription pays for reviews is a
  visible configuration decision rather than an edit to a secret nobody can read back.
- As the repository owner, I want both subscription tokens to live in the secret store at once, so
  that switching back costs a variable change instead of regenerating a token I overwrote.
- As a reader of a failed run, I want the annotation to name the secret the run actually looked
  for, so that a missing credential says which of the two is unset.

## Acceptance

- A repository variable `CLAUDE_CODE_PROFILE` set to `secondary` makes the workflow authenticate
  with `secrets.CLAUDE_CODE_OAUTH_TOKEN_SECONDARY`; any other value, and an unset variable, keep
  `secrets.CLAUDE_CODE_OAUTH_TOKEN`.
- The same credential the action spends is the one both leak checks compare the findings against,
  so a review can never be published unchecked against the token that produced it.
- The credential stays scoped to the steps that hold one today. Neither `decide` nor `announce`
  gains it, and no step that does not already carry it does.
- The two `is not set` annotations name the selected secret rather than a fixed one.
- A contract in `scripts/test-agent-workflow.sh` fails when the three places stop agreeing on which
  secret they resolve, and when the switch stops naming the secondary secret.
- `docs/operations/agent-workflow.md` describes the variable, both secrets, and what an unset
  secondary secret falls back to.

## Not in scope

Nothing in the repository reads a Claude credential outside `fathom-review.yml`; no other workflow
changes.

Krzysztof318:
`Fathom review` authenticated with one repository secret, `CLAUDE_CODE_OAUTH_TOKEN`, so spending a
second Claude subscription meant overwriting the token every run reads. Both tokens now live in the
secret store at once, and one repository variable selects between them.

## What changed

- `CLAUDE_CODE_PROFILE` set to `secondary` makes every run authenticate with
  `CLAUDE_CODE_OAUTH_TOKEN_SECONDARY`; unset, or set to anything else, keeps `CLAUDE_CODE_OAUTH_TOKEN`.
  It is a variable rather than a secret because the name of a profile is not one — which is what lets
  a run's log say which subscription it selected instead of `***`.
- The same condition resolves the credential in all three places the workflow holds one: the action
  input that spends it, and the two leak checks that refuse to publish findings containing it. A leak
  check comparing a review against the token the run never spent would match nothing, pass every
  review, and be green while doing it.
- The credential stays scoped to those three steps. Declaring it once in a workflow-level `env` was
  the alternative and would have put it in the environment of the two jobs that never touch one and of
  the step that runs the model, which is the opposite of what this file argues everywhere else. The
  workflow-level `env` carries the secret's *name* instead, so the two `is not set` annotations report
  the one the run actually looked for.
- `the_reviewer_resolves_one_claude_credential_everywhere` in `scripts/test-agent-workflow.sh` fails
  when the four spellings of the condition stop agreeing, and when a fourth step starts holding a
  Claude credential.

A `secondary` profile whose secret was never stored falls back to the first profile's token, because a
GitHub expression yields its right operand whenever the middle one is empty. A review running on the
other subscription beats one that cannot run, and both steps still refuse an unset credential outright
— both secrets missing is the case no fallback covers.

## Operator action

None until the second subscription is wanted. To switch: store `CLAUDE_CODE_OAUTH_TOKEN_SECONDARY`
from `claude setup-token`, then set the repository variable `CLAUDE_CODE_PROFILE` to `secondary`.
Switching back is deleting or changing that variable. No configuration key, database column, MCP tool
argument, or deployment contract moves, so nothing an operator runs is affected.

## Verification

`scripts/verify-full.sh` — passed, 7133 unit tests and 212 workflow contracts.

Closes #859

### DIFF
--- a/.github/workflows/fathom-review.yml
+++ b/.github/workflows/fathom-review.yml
@@ -111,9 +111,27 @@ permissions:
 #
 # It renders as nothing. A Markdown comment is invisible in the review GitHub displays and survives
 # the round trip through the API, which an invisible character or a trailing convention would not.
+#
+# Which Claude subscription pays for a review is the second thing declared here, and for a related
+# reason: three steps name the credential and they have to name the same one. `CLAUDE_CODE_PROFILE`
+# is a repository variable rather than a secret, because the name of a profile is not one — keeping
+# it out of the secret store is what lets a run say which subscription it selected instead of `***`.
+# Set it to `secondary` to spend the second profile's subscription; unset, or set to anything else,
+# means the first and `CLAUDE_CODE_OAUTH_TOKEN`. Both tokens live in the secret store at once, so
+# switching back is a variable change rather than a token regenerated over one that was overwritten.
+#
+# What is declared here is the secret's *name*, which the annotations that report a missing
+# credential print. The value is read per step, by the same condition, because a secret in a
+# workflow-level `env` is in the environment of every step of every job — including the two that
+# never touch a Claude credential and the one that runs the model — and this file's whole posture is
+# that the credential is held by the fewest steps that can do the work. `scripts/test-agent-workflow.sh`
+# fails when the four spellings of the condition stop agreeing, which is the one failure this split
+# can produce and the one nothing else would report: a leak check comparing findings against a token
+# no run spends passes every review it is given.
 env:
   AUTOMATIC_REVIEW_MARKER: '<!-- fathom-review: automatic -->'
   REQUESTED_REVIEW_MARKER: '<!-- fathom-review: requested -->'
+  CLAUDE_CREDENTIAL_SECRET: ${{ vars.CLAUDE_CODE_PROFILE == 'secondary' && 'CLAUDE_CODE_OAUTH_TOKEN_SECONDARY' || 'CLAUDE_CODE_OAUTH_TOKEN' }}
 
 jobs:
   decide:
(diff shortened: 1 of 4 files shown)


Krzysztof318:
The `Fathom review` check is red for an operational reason rather than a finding: the reviewer
stopped with `You've hit your weekly limit · resets Aug 18, 7pm (UTC)`, so no review was published
and no verdict exists for this head. Every other check on this head is green.

That limit is the thing this change exists to route around, and it cannot route around itself here:
`pull_request_target` runs the workflow file from `main`, so the profile switch this branch adds is
not in effect on its own pull request. Two ways to a verdict, both the owner's: merge this and then
store `CLAUDE_CODE_OAUTH_TOKEN_SECONDARY` and set `CLAUDE_CODE_PROFILE` to `secondary`, or wait for
the weekly window to reset and ask for a pass then.

PR #893, opened 2026-08-16
[#885] Model the authenticated From author as its own conclusion

### LINKED ISSUE #885: Distinguish authenticated transport identity from authenticated From author
## Context

#756 records what the receiving server authenticated from a trusted `Authentication-Results` header: DKIM signing domains, the SPF envelope domain, the DMARC result, and the domain displayed in RFC5322.From.

That evidence is necessary, but it does not by itself establish that the identity displayed in `From` is authentic.

`From` is attacker-controlled message content. A sender can write any address or domain there:

```text
From: billing@paypal.example
```

The presence of `paypal.example` in `From` is therefore never evidence that PayPal sent the message.

Likewise, a passing DKIM or SPF identity does not necessarily authenticate the identity displayed in `From`:

```text
From: billing@paypal.example
dkim=pass header.d=attacker.example
spf=pass smtp.mailfrom=attacker.example
dmarc=fail header.from=paypal.example
```

This message established that `attacker.example` authenticated. It did not establish that the displayed author `paypal.example` authenticated.

The distinction also matters for legitimate mail. A message may contain several passing DKIM signatures, use a sending subdomain, or be sent through a third-party delivery provider:

```text
From: user@example.com

dkim=pass header.d=provider.example
dkim=pass header.d=example.com
dmarc=pass header.from=example.com
```

Selecting one arbitrary passing DKIM identity is useful as authentication evidence, but it is not sufficient for deciding whether the displayed author authenticated.

DMARC is the receiving server's statement about whether an authenticated identifier aligned with the RFC5322.From domain according to the sender's published policy. MailFathom should consume that result rather than independently reconstruct DMARC alignment.

## Goal

Introduce a first-class conclusion about whether the identity displayed in RFC5322.From was authenticated.

This conclusion is separate from:

- the raw, attacker-controlled `From` value;
- DKIM signing identities;
- the SPF envelope identity;
- deployment-specific trust decisions made later by #757.

Conceptually:

```text
raw From
   |
   | attacker-controlled claim
   v
trusted Authentication-Results
   |
   +-- DKIM/SPF ----------------> authentication evidence
   |
   +-- DMARC + From ------------> author authentication
                                      |
                                      v
                              trust evaluation (#757)
```

The author-authentication result should distinguish:

```text
Authenticated
Failed
NotEstablished
```

### Authenticated

The trusted receiving server established that the identity displayed in `From` authenticated.

### Failed

The trusted receiving server evaluated authentication of the displayed author and reported that it did not hold.

### NotEstablished

The available trusted evidence is insufficient to conclude either success or failure.

`NotEstablished` is deliberately not equivalent to `Failed`.

### `From` is never trusted directly

The RFC5322.From domain remains untrusted message input until authentication evidence independently establishes it.

The implementation must never perform logic equivalent to:

```text
raw From domain
    |
    v
allow list / trusted domain check
```

Merely appearing in `From` cannot make an identity authenticated or trusted.

### DMARC

Where the trusted `Authentication-Results` header contains a usable DMARC result:

- `dmarc=pass` establishes the displayed RFC5322.From domain as authenticated;
- `dmarc=fail` records author authentication failure;
- `dmarc=none`, `temperror`, `permerror`, an unknown result, or no DMARC result do not establish successful author authentication.

MailFathom does not:

- resolve DNS for this decision;
- evaluate DMARC policy itself;
- calculate Organizational Domains;
- use a Public Suffix List to reconstruct relaxed alignment;
- reinterpret the receiving server's DMARC decision.

### Exact fallback when DMARC is unavailable

Where no usable DMARC result is available, a passing authenticated identity may establish the author only when its domain exactly matches the displayed `From` domain.

For example:

```text
From: user@example.com
dkim=pass header.d=example.com
```

may establish:

```text
AuthorAuthentication = Authenticated
AuthorDomain = example.com
```

Likewise:

```text
From: user@example.com
spf=pass smtp.mailfrom=bounce@example.com
```

may establish the same author domain.

A differing subdomain without a usable DMARC result does not allow MailFathom to infer relaxed alignment:

```text
From: user@example.com
dkim=pass header.d=mail.example.com
```

The result is:

```text
AuthorAuthentication = NotEstablished
```

not `Failed`, because MailFathom does not know whether the sender's DMARC policy would permit relaxed alignment.

### Multiple DKIM identities

All usable passing DKIM identities needed for author authentication must be considered.

One unrelated passing signature must not hide another passing signature that establishes the displayed author.

For example:

```text
From: user@example.com

dkim=pass header.d=provider.example
dkim=pass header.d=example.com
```

must allow `example.com` to be established as the authenticated author.

The first passing DKIM result may still be retained as evidence if useful, but it must not be the sole input to author-authentication decisions.

### Authentication evidence is not the author identity

It is valid for a message to contain:

```text
DkimDomain = attacker.example
AuthorDomain = paypal.example
AuthorAuthentication = Failed
```

This means the receiving server authenticated one identity while the identity displayed to the reader failed authentication.

That distinction must remain representable in the domain model.

## Existing `Alignment`

#756 currently records an exact comparison between the selected authenticated domain and the `From` domain.

That value must not be interpreted as DMARC alignment.

For example:

```text
From: example.com
dkim=pass header.d=mail.example.com
dmarc=pass header.from=example.com
```

may legitimately produce:

```text
Exact domain equality = false
DMARC = pass
AuthorAuthentication = Authenticated
```

The existing `Alignment` property should therefore either:

- be renamed to make explicit that it represents exact domain equality; or
- be removed if the new author-authentication result makes it redundant.

A future consumer must not be able to reasonably mistake it for the DMARC alignment result.

## Acceptance

- RFC5322.From is explicitly treated as attacker-controlled input and never trusted directly.
- The displayed author authentication result is modeled separately from raw `From`, DKIM identities, SPF identities, and deployment trust.
- Trusted `dmarc=pass` establishes the displayed author as authenticated.
- Trusted `dmarc=fail` records author authentication failure.
- MailFathom performs no independent DNS, SPF, DKIM, DMARC, Organizational Domain, or Public Suffix List evaluation for this decision.
- Without usable DMARC, an exact passing DKIM domain match may establish the author.
- Without usable DMARC, an exact passing SPF envelope-domain match may establish the author.
- Without usable DMARC, a differing subdomain produces `NotEstablished` rather than inferred relaxed alignment.
- Multiple passing DKIM results are considered such that an unrelated passing signature cannot hide another signature that establishes the displayed author.
- A message may retain passing DKIM or SPF evidence while its displayed author authentication is `Failed`.
- A message may have `dmarc=pass` while the exact domain comparison introduced by #756 is false.
- No allow-list or deployment-trust policy is evaluated against raw `From`.
- Tests cover:
  - exact DKIM/From match;
  - exact SPF/From match;
  - DMARC pass with a differing signing subdomain;
  - DMARC fail with an unrelated passing DKIM identity;
  - multiple passing DKIM signatures where only one matches `From`;
  - no usable DMARC plus a differing subdomain;
  - no usable `From` domain;
  - no trusted `Authentication-Results`;
  - a spoofed `From` domain with an authenticated unrelated transport identity.
- No address, domain, selector, or other sender identity is added to logs, metrics, or exception messages.

## Out of scope

- Deciding whether an authenticated author is trusted by this deployment — #757.
- Publishing the result to a caller — #888. Nothing renders a sender warning into a served body; #759 is closed as not planned.
- Independently verifying DKIM signatures.
- Independently evaluating SPF.
- Independently evaluating DMARC.
- ARC-based author authentication.

## Depends on

- #756 — provides the trusted authentication evidence this issue refines for downstream consumers.

## Follow-ups

- #757 must evaluate trust against an authenticated displayed-author identity, never against raw `From` and never against an arbitrary DKIM/SPF identity.
- #888 must publish an actual author-authentication failure as distinguishable from an authenticated but previously unknown correspondent.

Krzysztof318:
Closes #885

## What moves

`SenderAuthentication` recorded which identity the receiving mail server verified. That identity belongs to whoever
handed the message over, so a consumer reading it as a statement about the author a mail client displays would read a
relay's, a mailing list's, or a delivery provider's identity as the author's — and that reading is exactly what an
impersonation is built to get.

What the trusted evidence establishes about the **displayed author** is now a conclusion of its own, recorded beside the
identity it was reached from:

- **Authenticated** — a trusted `dmarc=pass`, or, where no usable DMARC result was reported, an authenticated identity
  whose domain is *exactly* the displayed one. The author domain is then the displayed domain rather than whichever
  identity established it.
- **Failed** — `dmarc=fail`, and nothing else. It ends the question rather than falling through to the exact
  comparison, because the receiving server reached it with the displayed domain's own published policy in hand.
- **Not established** — everything else, and deliberately not a failure: a signing subdomain with no DMARC result to
  interpret it, an identity belonging to somebody else, a message displaying no usable domain, no trusted header at
  all. Nothing here resolves DNS, computes an organizational domain, or consults a public suffix list, so relaxed
  alignment is never inferred.

**Every passing DKIM and SPF identity the trusted header reported is now considered**, rather than the first of each. A
message sent through a delivery provider carries the provider's verified signature beside its author's, and which one a
server lists first is the server's own ordering; reading only the first left ordinary mail with no established author.
The first is still kept as the recorded evidence, because the displayed author decides no part of what the verdict
records about the transport.

`SenderTrustPolicy` already judged the author rather than the raw `From` or an arbitrary identity, and it still reads
the same property; what changed underneath it is that the property is now a stored conclusion rather than a comparison
against one signature.

## Breaking change — database schema

`stored_emails` loses `SenderDomainAlignment` and gains `AuthorAuthenticationOutcome` (text, `NOT NULL`, default
`NotEstablished`) and `AuthenticatedAuthorDomain` (`character varying(253)`, nullable).

The dropped column held an exact comparison of the authenticated domain against the `From` domain. The author
conclusion is the answer a consumer actually wanted from it and accounts for DMARC, which that comparison could not;
the raw comparison stays derivable from the domains the row still carries. Keeping it — renamed — would leave a second,
weaker answer to the same question that a reader could still take for DMARC alignment, which it never was.

**Operator action:** apply the migration
(`20260816172328_ReplaceSenderDomainAlignmentWithAuthorAuthentication`) and nothing else. Rows already stored take the
`NotEstablished` default, which is what was true of them, and the
[extraction backfill](https://github.com/Krzysztof318/MailFathom/blob/main/docs/features/imap-synchronization.md) is
what re-derives the answer for mail stored earlier.

## Not done, and why

Steps 3 and 4 of `$add-migration` — applying the migration to the orchestrated database and proving the host starts
against the resulting schema — were **not performed**. The app model pins the PostgreSQL host port to `5432`, and
another session on this machine held it, so this checkout's orchestration proxied to that checkout's database instead
of its own. The migration was applied there by mistake and has been rolled back in full (its own `Down` plus the
`__EFMigrationsHistory` row), confirmed by reading the schema back. Step 2 — reviewing the migration as the SQL
PostgreSQL runs — was performed and is three statements: one `DROP COLUMN` and two `ADD COLUMN`, all metadata-only on
PostgreSQL 11 and above.

## Verification

- `scripts/verify-full.sh` — passed (212 contract checks, 7450 unit tests)
- `dotnet msbuild .config/CodeCoverage.proj -t:Collect` — 96.07% aggregate line coverage, threshold 85%
- `scripts/review-obligations.sh` — every page whose marker covers a changed path is in this change; no register moved

Tests cover the acceptance list on the issue: exact DKIM and SPF matches, `dmarc=pass` over a signing subdomain,
`dmarc=fail` over an unrelated passing signature, several passing signatures where only one matches `From`, a differing
subdomain with no usable DMARC, no usable `From` domain, no trusted `Authentication-Results`, and a spoofed `From`
beside an authenticated unrelated transport identity.


Krzysztof318:
The review reports that it did not reach the seven changed test files before answering, and asks for the acceptance
list to be checked against them in a follow-up pass. Here is that mapping, so the next reader has it without
re-deriving it. Every one of these ran in `scripts/verify-full.sh` (7450 passed, 0 failed).

| Acceptance item | Tests |
| --- | --- |
| exact DKIM/From match | `SenderAuthenticationReadingTests.Read_DkimDomainIsTheDisplayedOne_EstablishesTheAuthor`, `SenderAuthenticationTests.AuthorAuthentication_IdentityEqualToTheDisplayedDomain_IsAuthenticated` (`throughDkim: true`) |
| exact SPF/From match | `SenderAuthenticationReadingTests.Read_SpfDomainIsTheDisplayedOne_EstablishesTheAuthor`, the same theory at `throughDkim: false` |
| DMARC pass with a differing signing subdomain | `SenderAuthenticationReadingTests.Read_DmarcPassedOverASigningSubdomain_EstablishesTheDisplayedAuthor`, `SenderAuthenticationTests.AuthorAuthentication_DmarcPassOverAnIdentityThatIsNotTheDisplayedDomain_IsAuthenticated` (`mail.partner.test`) |
| DMARC fail with an unrelated passing DKIM identity | `SenderAuthenticationReadingTests.Read_DmarcFailedOverAnUnrelatedPassingSignature_RefusesTheAuthor`, `SenderAuthenticationTests.AuthorAuthentication_IdentityEqualToTheDisplayedDomainButDmarcFailed_IsFailed` |
| several passing DKIM signatures where only one matches `From` | `SenderAuthenticationReadingTests.Read_SeveralVerifiedSignatures_EstablishesTheAuthorFromTheMatchingOne`, `SenderAuthenticationTests.AuthorAuthentication_SeveralSignaturesWhereOneIsTheDisplayedDomain_IsAuthenticated` |
| no usable DMARC plus a differing subdomain | `SenderAuthenticationReadingTests.Read_SigningSubdomainWithoutDmarc_EstablishesNoAuthor`, `SenderAuthenticationTests.AuthorAuthentication_SigningSubdomainWithoutDmarc_IsNotEstablished` |
| no usable `From` domain | `SenderAuthenticationReadingTests.Read_MessageDisplayedNoUsableSender_EstablishesNoAuthor`, `SenderAuthenticationTests.AuthorAuthentication_NoDisplayedDomain_IsNotEstablished`, and `AuthorAuthentication_NoDisplayedDomainAndDmarcFailed_IsFailed` for the asymmetry |
| no trusted `Authentication-Results` | `SenderAuthenticationReadingTests.Read_NoHeaderFromTheTrustedServer_IsNotEstablishedRatherThanFailed`, `Read_MessageCarriedNoHeaders_IsNotEstablished`, `Read_AccountTrustsNoServer_IsNotEstablishedAlthoughAHeaderPassed` |
| spoofed `From` with an authenticated unrelated transport identity | `SenderAuthenticationReadingTests.Read_FromClaimsAnotherDomain_NamesTheAuthenticatedOneAndEstablishesNoAuthor`, `SenderAuthenticationTests.AuthorAuthentication_IdentityUnrelatedToTheDisplayedDomain_IsNotEstablished` (over all four DMARC results that decide nothing) |

Two more that are not on the list but are what the persistence half of the change owes:
`StoredEmailMetadataMappingTests.ApplyExtractedMetadata_EstablishedAuthor_RecordsTheAuthorBesideTheTransportIdentity`
proves the author reaches its own columns in the comparison form, and
`ApplyExtractedMetadata_NotEstablishedAfterAnAuthenticatedReading_ClearsTheWholeVerdict` proves a re-derivation clears
both new columns rather than leaving an author beside an identity a later reading disowned.

One thing the review could not have seen, stated here because it is the weakest part of this change: **steps 3 and 4 of
`$add-migration` were not performed**, for the reason the pull request body gives — the app model pins the PostgreSQL
host port, another session on this machine held it, and this checkout's orchestration therefore proxied to that
checkout's database. That misapplication has been rolled back in full and confirmed by reading the schema back. What
stands behind the migration here is step 2 alone: it was scripted and read as the SQL PostgreSQL runs, which is one
`DROP COLUMN` and two `ADD COLUMN` against `stored_emails`.

PR #1011, opened 2026-08-20
[#766] Group the MCP tools into categories a deployment publishes by

### LINKED ISSUE #766: Group the MCP tools into categories, published in full unless a deployment names the ones it wants
## Context

The MCP surface publishes five tools today and is on its way to publishing several times that: mailbox reads, sending, drafts, the outbox, contacts, the sender allow list, and rule authoring. Every one of them arrives with its own deployment switch and its own grant, which is right for what those control and wrong as the thing an operator uses to decide what their instance offers at all. Somebody standing up an instance for retrieval does not want to read seven settings to establish that it publishes nothing else.

A category is the coarse answer to that: tools are grouped by what they are for, every category is published by default, and a deployment may name the ones it wants. Naming none means the current behaviour, so an existing deployment is unaffected by the feature existing.

**Selection happens in two places, and only one of them is a control.** The deployment's configuration is the authority: it fixes what this endpoint may ever publish. A connecting client may then narrow further, by naming categories in a header on the requests that establish its session — which is what lets one deployment serve a retrieval-only agent and a full-capability one without running two endpoints, and what lets a client keep its own context small on a session where it knows it needs only part of the surface. A header is written by the caller, so it can only take away: a category the configuration did not publish stays unpublished however the header is written, and nothing in the header grants, enables, or reveals anything. That distinction has to be unmissable in the code and in the documentation, because a caller-supplied value that narrows is a convenience and the same value read as widening would be an authorization hole.

It stands beside the per-capability switches rather than replacing them, and the difference is worth stating because two mechanisms for one question is the failure mode this repository is careful about. A category says *this instance offers this kind of thing*; a capability switch says *this instance can do this at all* — sending is off until SMTP is configured and enabled, whatever any category says. So the category can only narrow: a tool appears when its capability is available and its category is selected, and no category selection turns anything on. That keeps one answer to *can this deployment send mail* while giving a separate answer to *what does this endpoint publish*.

There is a smaller reason too, and it is the one an agent's operator will feel first. A listing of thirty tools is thirty descriptors in a model's context on every session, and an instance used only for retrieval pays that for nothing.

## User stories

**As the operator of my own instance** I want to publish only the read tools from one endpoint, **so that** an agent I do not fully trust cannot see that anything else exists.

**As the operator of my own instance** I want a new release's tools not to appear on my endpoint without my noticing, **so that** what my instance publishes changes when I change it.

**As a client developer** I want a listing that holds only what this deployment offers, **so that** my agent's context is not spent on capabilities it cannot use.

**As a client developer** I want to ask for a narrower surface on a session of my own, **so that** one endpoint serves an agent that only reads and one that does everything.

## Acceptance

- Every published tool belongs to exactly one category, and the categories name what a tool is for rather than which project implements it. A tool without a category fails at startup rather than defaulting into one.
- Configuration selects categories. The unset value publishes every category, which is the behaviour a deployment has today, so the setting's absence changes nothing.
- Selection only narrows. A category naming a capability the deployment has not enabled publishes nothing, and no selection turns a capability on.
- A tool outside the selected categories is absent from `tools/list` and a call naming it is refused as an unknown tool, in the same shape and with the same coded failure a tool this deployment does not serve already produces.
- A connecting client may name categories in a header, and the effective set is the intersection of that header with the configured selection. A header can only narrow: a category the configuration excluded is never published because a header asked for it, and a test proves that directly.
- The header is untrusted input and is treated as such: it is bounded in length and in the number of values it may name, an unknown or malformed value is ignored rather than failing the request, and a header naming nothing usable leaves the configured selection in force rather than publishing an empty surface.
- The header's name, its syntax, and its narrowing-only semantics are documented where a client developer will read them, and the documentation says plainly that it is not an authorization mechanism.
- The chosen header does not collide with a header the MCP transport, the authentication methods, or a reverse proxy already uses, and the choice is stated.
- An unknown category name in configuration fails validation at startup, naming the value and listing what is accepted, rather than being ignored.
- The categories are a closed set, and adding one is a deliberate change to a named enumeration rather than a new string appearing in configuration.
- The documentation states the categories, which tools are in each, and that a category cannot enable anything.
- Unit tests assert that every published tool has a category, that the listing narrows as configured, that a call to an excluded tool is refused, that the unset setting publishes what is published today, and that a header naming an excluded category publishes nothing.
- The composed-host suite proves a narrowed listing over a real endpoint, with and without the header.

## Out of scope

Any per-caller authorization, which is the grant's question and #586's — the header narrows what a client asked for and decides nothing about what it may do. Any change to what a tool does. Any category that turns a capability on.

## Depends on

Nothing. It is worth landing before the surface grows rather than after.

Krzysztof318:
Every MCP tool now declares one category beside the permission it requires, and a deployment publishes the categories it wants instead of the whole surface.

## What a deployment writes

`McpEndpoint:PublishedToolCategories` names the categories this endpoint offers. Naming none publishes every one of them, which is what an endpoint that never had the setting has, so the key costs an existing deployment nothing. An unknown name fails validation naming the value and listing what is accepted, rather than narrowing an endpoint to a name no tool carries.

The six categories are `mailbox`, `flags`, `sending`, `drafts`, `answering`, and `contacts`. `send_draft` is in `sending` rather than in `drafts`, for the same reason it requires the sending grant: what it does is put mail on the wire, which is what lets a deployment publish `drafts` without `sending` and have an agent compose mail a person then reads.

The set is a closed enumeration, because the name is the identity and it travels outside the process. A tool registered without a category refuses to start the host that registered it — a startup gate over the registered tools rather than a decision taken per request.

## What a client may narrow

A connecting client may narrow its own session with the `MailFathom-Tool-Categories` header. The effective selection is the intersection with the configured one, so the header can only take away: a category the deployment excluded is never published because a request asked for it. The header is bounded in length and in how many names it may carry, an unusable value is dropped rather than refused, and a header naming nothing usable leaves the configured selection in force.

The name carries the product prefix so it collides with nothing the MCP transport, the authentication methods, or a reverse proxy defines, and it is added to the CORS allowlist so a browser client can send it.

## What a withheld tool looks like

Absent from `tools/list`, and a call naming it answered as a call naming a tool that does not exist — the same answer a caller outside its grant already receives, so nothing tells the two apart. A category enables nothing: a tool is served when its capability is available, its category is published, and the caller's grant reaches it.

## Breaking changes

None. The configuration key is additive and its absence is what every existing deployment has, so no operator action follows from this change.

## Verification

- `scripts/verify-full.sh` — passed (236 workflow assertions, 10183 unit tests, coverage gate)
- `scripts/build-docs-site.sh` — 0 errors, 4 pre-existing warnings
- `scripts/review-obligations.sh` — every row answered

Unit tests cover the enumeration, the selection and its narrowing, the header's bounds and parsing, the startup gate, and the composed filter pipeline with and without a header. `tests/IntegrationTests/Hosting/ComposedMcpToolCategoryTests.cs` proves the header over a real composed endpoint, listing and calling.

Closes #766


docs/operations/mcp-endpoint.md
```diff
@@ -719,6 +719,51 @@ an email address, a group, and a tenant claim are dropped at the boundary, so no
...
+## What this endpoint publishes
+
+A grant decides what an admitted caller may do. What this endpoint offers at all is a separate decision, and
+`McpEndpoint:PublishedToolCategories` is where it is written:
+
+```json
+{
+  "McpEndpoint": {
+    "Enabled": true,
+    "PublishedToolCategories": ["mailbox", "contacts"]
```
Krzysztof318, line 731: I assume default is enable all if not specified? Should be.

Krzysztof318: Yes — an unset key publishes every category, and so does an empty array. `PublishedToolCategorySelection.Of([])` returns `Everything`, which is what the host binds when the key is absent, so a deployment upgrading into this release sees no change to its endpoint and narrowing is always something an operator did deliberately.

Two tests hold that: `PublishedToolCategorySelectionTests.Of_NoCategoryAtAll_PublishesEverything` on the value, and `McpToolCategoryPublicationTests.ListTools_ADeploymentNamingNoCategory_IsServedEveryToolTheSurfacePublishes` through the composed filter pipeline.

The paragraph right under this block states it in bold, and the same sentence is in `docs/operations/configuration-endpoints.md` beside the key row.

Krzysztof318:
One correction to the record, on the candidate the review body raised and dismissed — the conclusion holds, the evidence cited for it does not.

`tests/IntegrationTests/Hosting/ComposedMcpToolCategoryTests.cs` is **not** part of `scripts/verify-full.sh`. The integration suite sets `IsTestingPlatformApplication` to `false` precisely so a solution-wide `dotnet test` never discovers it; it runs only through `scripts/run-integration-tests.sh` and the manually dispatched `Integration tests` workflow, so nothing in this pull request's green checks executed it.

What actually disproves the concern is `tests/Host.UnitTests/HostCompositionTests.cs`, which does run in `verify-full.sh`. It builds the real composed graph with `ValidateOnBuild` and `ValidateScopes` both on and then resolves `IHostedService`, so a hosted service capturing a scoped dependency fails there rather than on somebody's server. `PublishedToolCategoryGate` takes `IEnumerable<McpServerTool>`, and the SDK registers those as singletons, so there is no scoped capture to find.

PR #1076, opened 2026-08-21
[#1074] Move the .NET sources and tests under backend/ and open a frontend/ beside them

### LINKED ISSUE #1074: Move the .NET sources and tests under backend/ and open a frontend/ beside them
## Context

The repository root holds `src/` and `tests/`, and both contain nothing but the .NET solution's projects. A frontend project is coming, and it needs the same two directories for its own sources and tests. With the current layout there is no place to put them: a second `src/` cannot exist, and folding a frontend into the existing one would put TypeScript sources beside `Domain`, `Application`, and `Host` under a name that has meant "the .NET solution" everywhere it is referenced.

Nothing governs this beyond the layout itself. `docs/architecture/solution-structure.md` describes the tree as it stands, and the boundary rules in `AGENTS.md` are about the projects rather than about where the two directories sit, so the move changes no architectural decision — it only makes room for a second stack beside the first one.

The cost is that the two names are referenced from roughly 450 places outside them: the solution file, the container build, the coverage project, the docfx configuration, every verification and tooling script, every GitHub workflow, the `describes:` marker on most documentation pages, the skills under `.agents/skills/`, and the workflow contract itself. A move that misses one of those leaves a gate passing on a path that no longer exists.

## User stories

**As the repository owner** I want the .NET sources and tests under a single `backend/` directory, **so that** a frontend stack can take the same shape beside them without either one having to be told apart by project name.

**As a contributor** I want `frontend/src/` and `frontend/tests/` to already exist with the layout stated, **so that** the first frontend change adds code rather than deciding where code goes.

**As an agent working this repository** I want every reference to `src/` and `tests/` corrected in the same change, **so that** no script, workflow, or `describes:` marker silently points at a path that is gone.

## Acceptance

- `src/` and `tests/` are moved to `backend/src/` and `backend/tests/` with `git mv`, so history follows the files.
- `frontend/src/` and `frontend/tests/` exist and are tracked, carrying a `README.md` that states what each holds and that no frontend project has been added yet.
- `MailFathom.slnx` builds and every project reference, `Directory.Build.props` glob, shared-source include, and coverage path resolves against the new tree.
- Every script under `scripts/`, every workflow under `.github/workflows/`, `.github/fathom-review/`, `deploy/docker/Dockerfile`, `.config/CodeCoverage.proj`, `docfx/docfx.json`, and `aspire.config.json` names the new paths.
- Every `describes:` marker under `docs/`, and every prose path in `docs/`, `specs/`, `README.md`, `CONTRIBUTING.md`, `THIRD_PARTY_LICENSES.md`, and the skills under `.agents/skills/`, names the new paths.
- `AGENTS.md` and its per-directory files name the new locations, including the pointer table, and `src/AGENTS.md`, `tests/AGENTS.md`, `src/Infrastructure/AGENTS.md`, and `src/Mcp/AGENTS.md` move with the trees.
- `scripts/verify-full.sh` and `scripts/test-agent-workflow.sh` pass on the moved tree.
- `scripts/review-obligations.sh` resolves obligations against the new paths rather than reporting every changed file as uncovered.

## Out of scope

- Adding an actual frontend project, framework choice, package manager, or build. The directories and their README are the whole of the frontend half.
- Moving `tools/`, `MailFathom.slnx`, `Directory.Build.props`, `Directory.Packages.props`, `global.json`, or `NuGet.config` under `backend/`. The MSBuild infrastructure stays at the root, where it governs both `backend/` and anything added beside it.
- Renaming any project, assembly, root namespace, or the `MailFathom.*` prefix.
- Changing the CI matrix, the container image contract, or any published artifact path.

Krzysztof318:
## What this does

`src/` and `tests/` move to `backend/src/` and `backend/tests/`, and `frontend/src/` and `frontend/tests/` arrive beside them with a `.gitkeep` and a README each, and no project yet. The frontend work that comes next now has a decided place to land instead of having to decide one.

The build contract stays at the repository root — `MailFathom.slnx`, `Directory.Build.props`, `Directory.Packages.props`, `global.json`, `NuGet.config`, and `tools/`. MSBuild finds all of it by walking up from a project file, so a copy under `backend/` would govern nothing a root one does not, while a frontend build would still need its own manifest.

3092 files are pure renames. The rest are the references: the solution, `deploy/docker/Dockerfile` and its context allow-list, `.config/CodeCoverage.proj`, `docfx/docfx.json`, `aspire.config.json`, every script under `scripts/`, every workflow, `.github/fathom-review/`, the `describes:` marker on 64 documentation pages, the skills under `.agents/skills/`, and the workflow contract in `AGENTS.md`.

## Four references that were more than a path

- `.github/fathom-review/index-obligations.sh` derived the boundary name from the second path segment. That segment is now `src`, so every changed file would have been reported as owing tests in `backend/tests/src.UnitTests`. It reads the third.
- `scripts/test-agent-workflow.sh` created its fixture tree at `$repository_root/src` while the assertions beside it were rewritten to `backend/src`, so the suite failed at setup.
- The same suite's coverage-gap expectation was written in the order the unread paths sort in, and `backend/src/Other.cs` now sorts before `docs/features/sample.md`.
- `Protected paths` refused this pull request, and the second commit is why it no longer does. See below.

## Why `Protected paths` changes

The changed-files endpoint stops at 3000 entries, so the job refuses a larger pull request rather than clearing a question it could not compute. That is right for every author whose verdict the enumeration decides. The owner is not one of them: every protected path is theirs to change, so the list could only ever have named *which* ones they touched, and refusing them there withholds that list and blocks the merge over a verdict the list would not have altered. A tree-wide rename is the change that reaches it — it moves more files than the endpoint reports while touching a protected path only to correct a path written inside one.

The owner check moves ahead of the ceiling, and resolves once into `authored_by_owner` because both sites read it. The pass over the ceiling says the list is missing rather than passing in silence. Nothing changes for any other author: the refusal, the Dependabot exception, and the per-path annotations are untouched, and the contract test that covered the ceiling now covers it with an outside contributor as author, with a second test for the owner's case.

## Public surfaces

None breaks. The MCP tool contract and the configuration key set are unchanged — `PublicSurfaces.UnitTests` holds both golden files against the build and passes. The database schema is untouched and no migration is added. The deployment contract is unchanged: the image builds from the same context to the same entrypoint, and its `LICENSE`, `NOTICE`, and `org.opencontainers.image.licenses` are as they were. No package pin moves; every `packages.lock.json` is a pure rename.

## Documentation

Every `describes:` marker was rewritten and `scripts/test-agent-workflow.sh` resolves each one through git, so a marker naming nothing fails the suite. `docs/architecture/solution-structure.md` gains a paragraph on where the two stacks sit and why the build contract did not move with the solution, and its marker now covers `frontend/*/README.md`. `AGENTS.md` gains the same rule as its first architecture bullet, and `$get-started-contributors` names `frontend/` where it says where things live.

Five accepted ADRs have a path corrected in their body — 0001, 0003, 0004, 0005, 0006, and 0011. That is the one place this change edits a record that is closed, and it is the same category as the `describes:` marker the rules already exempt: a location the code lives at rather than anything about the decision.

## Verification

- `scripts/verify-full.sh` — passed on the rebased tree, digest recorded 2026-08-21T14:30:39Z.
- `dotnet test --solution MailFathom.slnx` — 10417 passed, 0 failed.
- `scripts/test-agent-workflow.sh` — 241 passed, 0 failed, including the two protected-paths ceiling contracts.
- `docker build --file deploy/docker/Dockerfile .` — image built. Nothing else covers the Dockerfile: `Container image` is manual dispatch only, so a broken build context would otherwise have surfaced at release time. Build context transferred 14.03 MB, so the `.dockerignore` allow-list still admits the sources and still keeps `bin/` and `obj/` out.
- `.github/fathom-review/index-obligations.sh` against a file list shaped as GitHub's — completed in 5m07s, one register row whose trigger and register both moved (`Directory.Packages.props` and `THIRD_PARTY_LICENSES.md`, both by path text alone), and 11 renamed sources with no referencing test, which a rename owes nothing for.
- `scripts/review-obligations.sh` did **not** complete. It synthesizes a patch for a rename, so on 3092 of them it builds a 29 MB added-lines corpus and greps 2600 type names over it; it was stopped twice after roughly twenty minutes. The direct index run above is what stands in for it, and `Fathom review`'s own indexing step takes the GitHub-shaped list rather than this one.

## Out of scope

Adding a frontend project, framework, or build. Moving `tools/`, the solution, or the MSBuild files under `backend/`. Renaming any project, assembly, or root namespace.

Closes #1074

- Issue: https://github.com/Krzysztof318/MailFathom/issues/1074
- Pull request: https://github.com/Krzysztof318/MailFathom/pull/1076


Krzysztof318:
Answering the one lead the first-pass review named and said it could not resolve, since nothing about it sits on a line and no thread carries it.

**`docs/operations/local-development.md` and `docs/operations/agent-workflow.md` are not stale.** Both were rewritten in this change. `local-development.md` carries 20 rewritten command lines — `dotnet run --project backend/src/AppHost/AppHost.csproj`, `dotnet user-secrets --project backend/src/Host/Host.csproj set …`, `aspire stop --apphost backend/src/AppHost/AppHost.csproj --non-interactive`, the benchmark run, and the coverage-scope prose — and `agent-workflow.md` carries the four `AGENTS.md` table rows plus its own `describes:` marker.

The review was right about *why* it could not tell: `changed_by_this_pull_request` in `obligations.json` is computed from the same truncated 100-file `files.json`, so it reads `false` for every page outside that window, including `docs/architecture/solution-structure.md`, which the body states was edited. The flag is unreliable past the first 100 files and the review should not have been able to confirm it either way.

What settles it without that flag is a repository-wide grep, which is also the check this change was built on. After the rebase onto `5c8fa200`:

```
git grep -n -P '(?<![A-Za-z0-9_./\\-])(?:\./)?(src|tests)/' -- ':!frontend/**'
```

returns exactly one line — the sentence in `AGENTS.md` that says the two stacks each carry *their own* `src/` and `tests/`, which is the one place the bare names are the point. Two narrower forms are checked beside it, for the shapes a path-boundary lookbehind cannot see: `(_root|/head)/(src|tests)` for a shell variable followed by the directory, and `\\[ntr](src|tests)/` for a path inside a `$'…'` escape. Both are empty. Those two forms are not pedantry — each hid a real defect earlier in this change, and both are named in the body.

**On coverage.** The verdict is an approval over the first 100 of 3232 files, and it is worth exactly that. The change is 3092 pure renames, which git reports at 100% similarity and which no reader has to open, and 140 files carrying content — 111 modified, 29 renames with an edit inside them. The three gates that did read all of it are `scripts/verify-full.sh`, the 241-contract suite, and a local `docker build`, and what each one proves is in the body.

Issue #1398, opened 2026-08-31
Write the client's contract family: quality, UX, UI, accessibility, and performance


Krzysztof318:
## Context

`frontend/src/AGENTS.md` was 565 lines of contract that every agent session touching the client loaded before reading a file: what the application is, MVUX as the state model, the XAML-first rule and where code-behind was correct, where styles and brushes live, the responsive and native-feel obligations, which awaits keep the synchronization context, the project and build layout, what a published head is trimmed with, and how the client reaches the backend. #1392 removes it, because almost every line of it decided something about Uno.

The obligations underneath it did not go anywhere. A React client still has a state model, still has a rule about where markup ends and logic begins, still has styling that belongs somewhere rather than anywhere, still has a responsive obligation, and still reaches the backend over `/api/client`. This writes that contract for the stack that replaced Uno.

**The family mirrors the service stack's**, which #1406 is rebalancing at the same time. `frontend/AGENTS.md` holds what is true across the whole stack, `frontend/src/AGENTS.md` what governs the application source, and `frontend/tests/AGENTS.md` what governs the suite — the last of those written by #1397, which this issue does not duplicate.

**Write it the way the service stack's files are written.** They state a decision and the reason for it, they close a question so it is not re-answered per file, they say what a tool already enforces and stop rather than restating it, and they name a worked example in the repository instead of describing one. A rule that could be inferred from the code is not worth a line; a rule a competent developer would get wrong half the time is worth a paragraph.

**Code quality is the first emphasis and it is not the linter's job.** The linter and the compiler own everything they can see, which #1396 sets restrictively. What is left is what a tool cannot decide: where a component's boundary is, when state is derived rather than stored, why an effect is almost always the wrong answer, how far a prop may travel before the shape is wrong, what a module may know about another, and what a name has to communicate. Those are the rules that decide whether the client is readable at twenty screens.

**UX design is the second, and it is a contract rather than taste.** The Uno file carried a rule under the heading *Nothing waits in silence, and nothing waits without a way out*, and that rule is stack-independent and worth keeping by name: every surface that waits shows that it is waiting, every failure says what failed and offers the way out, and no state is reachable that a person cannot leave. Beside it belong the states a screen must actually have — loading, empty, partial, error, and offline — what a destructive action must confirm, where focus goes when a view changes, and what may never shift under a reader's cursor.

[...]

Issue #1404, opened 2026-08-31
Drive and check the client in a real browser with the agent Playwright CLI


Krzysztof318:
## Context

The Uno client could not be inspected from outside itself. Its Skia renderer drew to a canvas inside an `iframe`, so a browser saw two nodes and no more until an off-screen skip link was activated by keyboard, and every interaction before that was a pixel coordinate. Verifying a screen meant the App MCP reaching into a running application's visual tree, which is a different thing from what a person's browser does.

A React client is ordinary DOM. Roles, names, and labels are real, a locator resolves the way it resolves on any page, and the network panel shows exactly what the client sent and what came back. That makes browser-driven verification cheap here in a way it was never cheap for Uno, and it is why this is a child of the foundation rather than something deferred until the client has screens worth checking.

**The headline is the agent-oriented CLI, `@playwright/cli`.** This is Playwright's newest command-line tool and it is built for agents rather than for a person at a terminal or for a test runner — it is a different package from `@playwright/test` and the two are not interchangeable. It holds a browser open across invocations as a named session, takes an accessibility snapshot of the page and hands back stable element refs, and then acts on those refs: `snapshot`, `find`, `click`, `fill`, `press`, `generate-locator`. Crucially it also opens the parts of a browser that are otherwise hand-written — `requests` numbers everything the page issued, `request` and `response-body` open one of them, `console` reports what the app logged, and `tracing-start` and `video-start` record a run to look at afterwards. It ships its own agent skill inside the package, which is the current reference for how to drive it.

**This is what makes agent-written UI verifiable at all.** An agent that writes a screen and cannot open it is guessing, and a screenshot pasted into a reply is not evidence of behaviour. With this CLI the loop closes: build, open, snapshot, interact, read the request that actually went out, and correct — before anything is committed, rather than after a review asks. The repository already treats a claim without evidence as a claim, and this is the instrument that produces the evidence for client work. Expect it to be used heavily and continuously, not occasionally.

**Two uses, two packages, one browser engine.** They are not the same activity and conflating them is the mistake this issue exists to prevent:

- **`@playwright/cli` — an agent driving the running client.** Development, diagnosis, and self-verification. Not a gate, not committed as tests, used constantly.
[...]

PR #1431, opened 2026-08-31
[#1409] Localize the client in English and Polish, with English as the default

### LINKED ISSUE #1409: Localize the client in English and Polish, with English as the default
## Context

The Uno client was localized in English and Polish by #1108, which shipped in `0.8.0` and went with the client #1392 withdrew. Nothing under `frontend/` carries a string catalogue, a locale, or a locale-aware formatter today, and the children of #1394 rebuild the foundation without one: the delivery table names the workspace, the test suite, the contract family, the desktop shell, the image, the browser harness, the CI job, and the licence register, and no row names language.

The reason #1108 gave for doing this before the screens holds unchanged on the new stack, and it is the whole argument for a row in wave 2 rather than an issue somewhere after the client roadmap. Retrofitting a string catalogue across pages already written is the expensive version of this work, and every screen added first is another screen to convert. #1178 and the stage parents beneath it are re-read against React separately, and each of those screens is written against whatever this issue leaves in place — so this is the difference between a screen written with its strings outside it and a sweep through every screen later.

`docs/features/presentation-plan.md` already states the client is localized, and reasons at length about a column heading being "words in somebody's language" and about what shipping the English word to a Polish reader would cost. That is a claim about the client the tree currently cannot honour.

**What the platform gives free is worth stating, because it decides how small the addition is.** `Intl` formats dates, numbers, relative times, lists, and plural categories in every engine both heads run on, so what a library would be taken for is message catalogues, interpolation, and plural selection on top of `Intl.PluralRules` — never formatting. Whatever is chosen is weighed against writing that much by hand, in the order the repository's own rules put it: platform capability first, then a package with a clear owner and purpose.

**One property changed with the stack.** Uno's `ILocalizationService.SetCurrentCultureAsync` applied a culture on restart, so #1108 had to tell the person choosing that the switch was not live. React re-renders, so that constraint does not survive the platform change and the acceptance below says the switch applies immediately rather than carrying the old caveat forward.

## User stories

**As** somebody using MailFathom in Poland **I want** the client in Polish, **so that** I read my own mail in my own language.

**As** somebody anywhere else **I want** English without configuring anything, **so that** a first run is readable before I have found any setting.

**As** a client developer **I want** every visible string to live outside the component that shows it, **so that** the third language is a catalogue rather than a sweep through every screen.

## Acceptance

- `Client.App` resolves every user-visible string through one localization mechanism, `en` and `pl` each carry a catalogue complete for what the client shows today, and `en` is the default and the fallback for a key or a locale that resolves to nothing.
- A user-visible literal written into a component fails `pnpm lint` rather than being caught in review, and the check is demonstrated by the error it produces rather than asserted.
- The locale of a first run is read from the browser or operating-system preference, narrowed to what the catalogues carry, falling back to `en`. An explicit choice overrides it and survives a restart of either head.
- Changing the language applies without a restart, on both the web and the desktop head.
- Dates, numbers, and relative times are formatted through `Intl` under the active locale, rather than through a second copy of that formatting inside the catalogues.
- `Client.Backend` carries no catalogue, no locale, and no localization dependency. The boundary #1396 made a resolver decision holds here too: language is a presentation concern and lives in `Client.App`.
- The suite asserts that the offered locales are the configured ones, that a key present in one catalogue and missing from the other fails, and that an unknown locale resolves to `en` — in whichever form #1397 settles for the client's tests.
- `.config/typos.toml` excludes the Polish catalogue. Its exclusion list governs the walk rather than a path named on the command line, so the entry is written and then confirmed by running the check rather than assumed.
- `THIRD_PARTY_LICENSES.md` records whatever package this introduces. Anything reaching the bundle is a redistributed component under ADR 0016 and is recorded as latent while no artifact carries it, exactly as React is today.

## Out of scope

- **Regional variants.** `en` and `pl` are neutral locales; `en-GB` and `pl-PL` arrive when something actually differs between regions, and nothing does yet.
- **Right-to-left layout**, and any language beyond the two.
- **Localizing what the service returns.** Every message the backend produces is a surface of its own with its own reasoning, and asking for one through `Accept-Language` on `/api/client` is a change to that contract rather than to the client.
- **Documentation, the README, and anything outside `frontend/`.** The repository's own rule that every artifact in it is written in English is untouched by the client speaking two languages.
- **Any actual screen.** The catalogues carry what the shell and the placeholder screen show today; a screen arriving later brings its own keys.

## Dependencies

#1396 stands up the workspace, the two packages, and the toolchain a catalogue and its lint rule are added to. Nothing here can start before it.

#1397 states the client's test contract and runs beside this. Where it has landed first, the assertions above are written in the form it settles; where it has not, they arrive with it rather than being skipped.

#1398 writes the client's contract family, and the rule that no component carries a user-visible literal belongs in it. The two have no order against each other: whichever lands second states the rule against the mechanism the other one built.

#1108 is the Uno-era version of this work. It closed as completed and was withdrawn with the client by #1392, so it is the record of what was decided about language rather than a contract to restore — its acceptance is written against `x:Uid`, `.resw`, and `UnoFeatures`, none of which exists here.

Krzysztof318:
Closes #1409

`Client.App` now reads in English and in Polish. English is the default and the fallback; a first run
with no choice stored reads what the browser or the operating system prefers, narrowed to a language a
catalogue was written for; the control in the header overrides that, the choice survives a restart of
either head, and changing it rewrites the screen without anything restarting.

## No library, and why

The issue asked for the addition to be weighed against writing it by hand, in the order the repository's
rules put it. It was, and nothing was added: **no package, no bundle weight, and no row in
`THIRD_PARTY_LICENSES.md`.**

`Intl` already formats dates, numbers, relative times, lists, and plural categories in every engine both
heads render in, so what a library would have been taken for is a catalogue, a lookup, and a `{name}`
hole to fill — 160 lines across five files, less than the configuration `i18next` or FormatJS is adopted
with, and it buys something neither gives for free: `pl.ts` is annotated with the type `en.ts` exports, so
**a key one language carries and the other does not fails `pnpm typecheck`** rather than reaching a screen
as a missing string. The suite asserts the same parity at run time as well.

## What is here

- `localization/en.ts` declares the keys and is the fallback; `localization/pl.ts` is typed against it.
- `localization/locale.ts` resolves which language a run opens in — the stored choice, else
  `navigator.languages` narrowed to what is carried, else `en` — and stores an explicit one. A stored
  value naming a language the client does not carry is refused rather than trusted.
- `localization/Localization.tsx` is the provider; `localization/useLocalization.ts` is the context and the
  hook. The two are separate files because a module Vite hot-reloads may export components alone, which
  `react-refresh/only-export-components` states.
- The provider also keeps `document.documentElement.lang` in step with the active language, which is what
  assistive technology pronounces by.
- `App.tsx` resolves every word through the catalogue, and formats the last-synchronized instant with
  `Intl.DateTimeFormat` under the active locale. The two closed sets `Client.Backend` answers with —
  `MailSynchronizationState` and `ClientFailureReason` — are mapped to message keys by two tables that are
  exhaustive by their own type, so a value added to either fails to compile until this screen names it.

## A user-visible literal fails `pnpm lint`

Four `no-restricted-syntax` selectors in `eslint.config.ts`, which needs no plugin and adds nothing to the
bundle. Demonstrated by writing the offending lines and running the command rather than asserted:

```
src/Client.App/src/literalProbe.tsx
  3:29  error  This attribute is read out to somebody, so it is a user-visible string: put it in src/Client.App/src/localization/en.ts and pass translate() to it   no-restricted-syntax
  4:16  error  A user-visible string belongs in src/Client.App/src/localization/en.ts, with its Polish counterpart, and reaches the screen through translate()      no-restricted-syntax
  5:17  error  A user-visible string belongs in src/Client.App/src/localization/en.ts and reaches the screen through translate(); a number is written with Intl …   no-restricted-syntax
  6:17  error  A sentence assembled in markup cannot be reordered by a translator. Write it as one catalogue entry with a {name} hole and fill it through translate() no-restricted-syntax
```

A second probe confirmed it stays quiet on what it must not catch: `className`, `data-*`, `tabIndex={0}`,
`type="text"`, a form `defaultValue`, and whitespace-only JSX text. Both probes were deleted; neither is in
the diff.

## Three acceptance items answered rather than met literally

- **Plural selection is not shipped, because nothing counts.** The catalogues carry what the shell and the
  placeholder screen show today, and no screen displays a count. `Intl.PluralRules` is what selects a form
  when one first does — Polish has four categories and English hides that — and
  `frontend/src/AGENTS.md` now says so where the screen that needs it will be written. Adding the machinery
  now would be a mechanism with no caller; adding it later touches one module rather than every screen,
  which is the retrofit cost this issue exists to avoid and is not one of them.
- **A number and a relative time have no caller either.** The rule the acceptance states is that formatting
  goes through `Intl` under the active locale rather than through a second copy inside the catalogues, and
  that rule is now stated in the contract and held by the one value the screen does format. No catalogue
  carries a month name, a separator, or a hand-written relative time.
- **There is no runtime fallback for a missing key**, because the type system makes one unreachable, which is
  the stronger form of what was asked. English remains the fallback for a *locale* that resolves to nothing.

## The typo check reads `pl.ts` on this pull request anyway

`.config/typos.toml` now excludes the Polish catalogue, and the entry governs the walk `typos` does over a
directory. It does not survive a path being named on the command line, which is exactly what
`.github/workflows/typo-check.yml` does with the files a pull request changed — so this pull request's own
`Typo check` reads `pl.ts`. Confirmed clean by running the pinned binary rather than assuming:
`typos v1.48.0 --config .config/typos.toml` over every changed path, exit 0.

## An environment fact the suite needed

`localStorage` reads as `undefined` under Vitest here, and the cause is not jsdom: Node publishes a Web
Storage implementation of its own, the jsdom window is the worker's global object, and Node's getter is
therefore the one on it — answering `undefined` unless the process was started with `--localstorage-file`.
jsdom's own storage is present under an internal name; `vitest.setup.ts` puts it back under the right one.
`frontend/tests/AGENTS.md` records it so the next person does not rediscover it.

## Against the contract family

Rebased onto #1398, which landed while this was being written. The rule that no component carries a
user-visible literal is enforced by lint, so per `frontend/AGENTS.md` it is *not* restated in an
`AGENTS.md`; `frontend/README.md` carries the mechanism and how to reproduce each refusal.
`frontend/src/AGENTS.md` gains the half a linter cannot decide — a string is a catalogue entry however it
reaches the screen, a sentence is one entry with holes rather than fragments joined at the call site,
nothing branches on the language, and no catalogue or locale crosses into `Client.Backend`, which is now the
fourth item on that file's list of things that may never cross.

## Not addressed here, deliberately

The placeholder screen's failure copy is one sentence with the reason interpolated, which does not meet the
`frontend/src/AGENTS.md` rule that four failure reasons are four sentences and four next steps. That rule
arrived with #1398 and is about the screen rather than about its language; this change translates what is
there rather than redesigning it, since the issue puts any actual screen out of scope.

## Verification

- `scripts/verify-full.sh` — passed, 271 contract tests, client stack resolved and run.
- `pnpm typecheck`, `pnpm lint`, `pnpm test` (56 tests, 7 files), `pnpm format:check` — all clean.
- `scripts/review-obligations.sh` — every documentation row confirmed in the file it points at.
- `$check-docs-licenses` — Docs pass, Changelog n/a, Licenses n/a (no package, tool, service, or image
  introduced; `frontend/package.json` and `pnpm-lock.yaml` are untouched).

No public surface breaks: the MCP tool contract, the configuration schema, the database schema, and the
deployment contract are all unchanged.


frontend/src/Client.App/src/localization/Localization.test.tsx
Krzysztof318, line 1: why here? not in tests fdir?

Krzysztof318: Because `frontend/tests/AGENTS.md` puts every client unit test beside the source it covers, and the reason is the package boundary rather than taste.

`frontend/tests/` is outside both packages. A test written there could reach `Localization.tsx` only through a relative path out of `Client.App`, which is the one thing that boundary exists to refuse — and it would resolve neither package's manifest, so a `Client.Backend` test there could import React while its source cannot. A test *inside* the package inherits the boundary instead, and nothing has to check that it did not.

So this file follows what is already there — `App.test.tsx`, `mailAccounts.test.ts`, `failure.test.ts`, `session.test.ts` all sit beside their source — rather than introducing a placement of its own.

What does live in `frontend/tests/` is the contract itself and, since #1404, the browser suite: that one drives the built bundle over a preview server instead of importing a module, so it belongs to neither package and has nothing to resolve. This change adds one test there too, for the single claim jsdom structurally cannot make — that a language chosen in one load is still the language after a real reload.

frontend/eslint.config.ts
```diff
@@ -69,5 +69,39 @@ export default defineConfig(
...
+                        'A user-visible string belongs in src/Client.App/src/localization/en.ts and reaches the screen through translate(); a number is written with Intl under the active locale.',
+                },
+                {
+                    selector: ':matches(JSXElement, JSXFragment) > JSXExpressionContainer > TemplateLiteral',
+                    message:
+                        'A sentence assembled in markup cannot be reordered by a translator. Write it as one catalogue entry with a {name} hole and fill it through translate().',
+                },
+                {
+                    selector:
+                        'JSXAttribute[name.name=/^(alt|title|placeholder|aria-label|aria-description|aria-placeholder|aria-roledescription|aria-valuetext)$/] > Literal',
```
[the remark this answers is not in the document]
Krzysztof318: Fixed in 36b5fb11. The gap was real: `alt={'text'}` sat one level deeper than the selector reached, and none of the other three matched it because they anchor on `JSXElement`/`JSXFragment` children.

The attribute list is now written once as a constant and matched in all three written forms — `> Literal`, `> JSXExpressionContainer > Literal[value=/\S/]`, and `> JSXExpressionContainer > TemplateLiteral` — so the list cannot drift between them.

Deliberately a two-level step rather than a descendant match: `aria-label={translate('shell.language')}` carries a string literal too, and it is the catalogue key rather than a sentence. A descendant selector would refuse the correct form along with the wrong ones. The residual gap that leaves is a literal inside a larger expression, `aria-label={ready ? 'a' : 'b'}`, which no selector can separate from the key case; it is narrower than what was reported and is named here rather than left silent.

Demonstrated rather than asserted. A probe with all three forms:

```
   8:22  error  This attribute is read out to somebody, so it is a user-visible string: …  no-restricted-syntax
   9:23  error  This attribute is read out to somebody, so it is a user-visible string: …  no-restricted-syntax
  10:23  error  This attribute is read out to somebody, so it is a user-visible string: …  no-restricted-syntax
```

lines 8, 9 and 10 being `alt="…"`, `alt={'…'}` and ``alt={`…${x}`}``. The same probe left `aria-label={translate('shell.language')}`, `className`, `data-*`, `tabIndex={0}`, `type="text"` and a form `defaultValue` unreported. The probe is deleted and is not in the diff; `frontend/README.md` gained the braced row so the documented mechanism matches the enforced one.

Issue #1397, opened 2026-08-31
Test both client packages with Vitest, and state the contract that governs it


Krzysztof318:
## Context

The Uno client had a unit suite on xUnit.net v3 and a `frontend/tests/AGENTS.md` stating its policy — what a test asserts, how a feed is asserted, what belongs in the suite and what does not, and what runs it. #1392 removes both, and the root `AGENTS.md` rule that unit tests are part of every behaviour change in both stacks survives that removal without a suite to satisfy it.

This restores the suite and the contract on the React stack, with Vitest and React Testing Library. Vitest shares the application's resolver and TypeScript configuration rather than needing a second one, which is most of why it is the runner here.

**Both halves of the client are tested, not one.** The two packages #1396 establishes are tested differently and both properly:

- `Client.Backend` is ordinary logic with no DOM in its closure — request construction, response parsing, the failure model, the session. It is the half that is cheap to test thoroughly and there is no excuse for thin coverage of it.
- `Client.App` is components, state, and screens. They are tested through what a person sees and does — rendered output and interaction — rather than through internal state, and as far as jsdom can reach. Where a component genuinely cannot be asserted without a real browser, that is what the Playwright harness in #1404 is for, and the test moves there rather than being dropped.

"As far as possible" is the standard on the second half and it is meant as a demand: a component with behaviour and no test is a gap to justify, not a default.

**`backend/tests/AGENTS.md` is the model to write against.** It is 115 lines that decide things — the unit testing policy, what properties and examples are for, how coverage is read, what a cost claim may assert, and what belongs in the integration suite instead — rather than describing a test framework. The client's file is written the same way: a decision and its reason, closing a question so it is not re-answered per test, and silent about anything the runner already enforces. It is one member of the contract family #1398 establishes, and this issue writes it.

**The policy is the part worth reviewing, not the runner.** The root `AGENTS.md` rules that hold whatever the language is — one test describes one behaviour, tests are fast, isolated, repeatable, order-independent, parallel-safe, and never touch a real clock, network, database, or the filesystem — need their React expression: what stands in for `TimeProvider` when a component reads the current time, what is faked at the network boundary, and what an assertion about a rendered screen is allowed to reach for.

**Two rules the Uno contract carried do not survive the stack** and need replacing rather than translating: `Member_Scenario_ExpectedBehavior` naming is a C# convention, and asserting an MVUX feed has no counterpart.

## User stories

[...]

PR #1443, opened 2026-08-31
[#1441] Collect coverage of the client's unit suite, reported and never enforced

### LINKED ISSUE #1441: Collect code coverage of the client's unit suite, reported and never enforced
## Context

The client's unit suite has run under Vitest since #1397, and nothing measures it. `frontend/tests/AGENTS.md` § *Coverage* records that as a decision deliberately deferred — "what a number over this stack would be worth, is a decision of its own" — and `docs/operations/agent-workflow.md` states the full gate runs the client flow "with no coverage threshold because there is no client code measured yet". The service stack has had the opposite arrangement since it existed: `.config/CodeCoverage.proj` collects, merges, and enforces 85% aggregate line coverage.

The owner has now taken that decision: collect coverage of the client suite. The half the repository has already answered twice is whether to gate on it. It does not. `scripts/run-integration-tests.sh` collects an integration report and enforces nothing — "Unit tests stay the only source of an enforced coverage metric" — and `docs/operations/agent-workflow.md` § *The mutation score is read, never enforced* records why the service's own 85% figure stopped discriminating: it reports that a line ran rather than that anything asserted its result. A second enforced number over a suite this young would inherit that failure before it had earned anything, so this collects and reports and gates on nothing.

Nothing else governs this. No ADR covers either stack's coverage arrangement, and none is opened for this one: what a collector is configured to measure is a tooling decision recorded where the tooling is.

## User stories

**As the owner** I want to see which of the client's modules the unit suite actually reaches, **so that** a screen or a request builder nobody tested is visible as a number rather than only as an absence somebody would have to notice.

**As an agent implementing a client change** I want the coverage of what I wrote reported by the command I already run, **so that** a module I left untested shows up in the same pass as a lint error rather than in a review round.

**As a contributor** I want the client's coverage to gate nothing, **so that** a legitimately untestable module does not have to be argued past a threshold measuring execution rather than assertion.

## Acceptance

- `@vitest/coverage-v8` is pinned at the version matching the `vitest` pin in `frontend/package.json`, `frontend/pnpm-lock.yaml` is regenerated by pnpm, and `THIRD_PARTY_LICENSES.md` carries it — the row it belongs in plus the JavaScript census, which nothing recomputes.
- `frontend/vitest.config.ts` collects coverage over both Vitest projects on every `pnpm test`, with no second invocation beside it: the contract in `frontend/tests/AGENTS.md` is that `pnpm test` is the whole of how the suite runs, so a `test:coverage` script would be the drift that rule refuses.
- The report is written under the repository-root `artifacts/` directory, which `.gitignore` already covers, and a summary is printed where a gate run and a CI job both show it.
- No threshold is configured, and neither verification script nor any workflow fails on the figure.
- `frontend/tests/AGENTS.md` § *Coverage* states what is collected, what is excluded and why, and that nothing is enforced — replacing the paragraph that defers the decision. `docs/operations/agent-workflow.md` and `docs/operations/local-development.md` are corrected wherever they say the client measures nothing.
- The client suite passes and both verification gates are green.

## Out of scope

- Any threshold, gate, or required check over the client figure, in either verification script or any workflow.
- Uploading the client report as a CI artifact, as `build-test-format-and-migrations.yml` does for the service. The summary in the job log is what this issue delivers; an artifact is a separate decision about what is worth retaining.
- The browser suite under `frontend/tests/`. It drives a built bundle over HTTP rather than importing a module, so a provider instrumenting the module graph does not reach it.
- The desktop shell's Rust crate under `frontend/src-tauri/`. Its own tooling is `cargo`, and neither verification gate reaches the crate graph at all.
- Revisiting the service's 85% threshold or the mutation score.

## Dependencies

None.

Krzysztof318:
Closes #1441

## What changes

The client's unit suite has run unmeasured since #1397. `frontend/tests/AGENTS.md` § *Coverage* recorded that as a decision deliberately deferred rather than as a step nobody got to, so this change takes the decision and writes down both halves of it: coverage is collected, and nothing is enforced on it.

**`@vitest/coverage-v8` 4.1.11**, pinned at the runner's own version because its `peerDependencies` require exactly that, and `frontend/vitest.config.ts` turns the provider on for both Vitest projects. Three things there were decided rather than derived:

- **It runs on every `pnpm test` rather than behind a flag or a second script.** `frontend/tests/AGENTS.md` states that `pnpm test` is the whole of how this suite runs, locally and in CI, "and there is no second invocation to drift from it" — a `test:coverage` script would be that drift. The v8 provider is what makes it affordable: it reads counters the runtime already keeps rather than instrumenting the module graph, so the suite's wall clock does not move.
- **The measured scope is both packages' `src/` whether or not a test imported the file.** Without an `include` the report would name only what a test happened to reach, and the module nobody covers — the thing the number exists to show — would be absent rather than sitting at zero. `stubMailFathom.ts` is that file today.
- **Two exclusions**: declaration files, which state types and run nothing, and `Client.App/src/main.tsx`, which mounts React into the document and decides nothing — the client's counterpart to `Host` and `AppHost`, left out of the service's denominator for the same reason. Vitest drops the suite's own test files.

**No threshold, in either verification script or any workflow**, and the service's 85% stays the repository's only coverage gate. The reasoning is already in the tree and this follows it rather than inventing a second one: `docs/operations/agent-workflow.md` § *The mutation score is read, never enforced* records that the service's figure has sat above 95% for months and reports that a line ran rather than that anything asserted its result, and `scripts/run-integration-tests.sh` collects an integration report under the same rule. A second enforced number over a suite this size would inherit that failure before earning anything.

The report goes to the repository-root `artifacts/coverage/client/`, beside the service's and inside the directory `.gitignore` already covers, so the full gate's untracked-file check is unaffected — verified by running the suite and reading `git status --untracked-files=all`.

Documentation moved with it: `frontend/tests/AGENTS.md` § *Coverage* replaces the deferral, `docs/operations/local-development.md` gains § *The client's suite* under the coverage section and opens that section with the two-stack split it never had, `docs/operations/agent-workflow.md` loses the sentence that said no client code is measured, and `frontend/README.md` gets one line.

## How it was verified

`bash scripts/verify-full.sh` on this branch, rebased onto current `main` — see the checklist below.

- The client suite: 7 files, 56 tests, green, with the collector on. Coverage reads 95.04% statements, 92.75% branches, 97.14% functions over 121 statements. That is a first reading rather than a baseline anything holds to, and it gates nothing.
- Checked by hand that the report lands under the ignored `artifacts/` and leaves the working tree clean, and that the collector's own configuration behaves as documented: with `coverage.exclude` set the way a first attempt sets it, Vitest replaces its default list rather than extending it — the defaults are empty in Vitest 4 and `include` is the whole control, which is why the exclusions here are written out.
- `bash scripts/review-obligations.sh`: one register obligation, satisfied in this change. The pages it named as documentation candidates were read; none carried a sentence this change makes false beyond the ones corrected here.
- `THIRD_PARTY_LICENSES.md` § *The client's two dependency closures* was re-enumerated with the commands that section publishes rather than reasoned about: 238 packages at 246 name-version pairs to **249 at 258**, MIT 184 → 192, BSD-3-Clause 3 → 6. The eleven packages added are the provider, `ast-v8-to-istanbul`, `@bcoe/v8-coverage`, `magicast`, `make-dir`, `html-escaper`, `has-flag`, `supports-color`, and the three istanbul reporters it remaps through; `js-tokens` gained a second version. Nothing was removed and no new licence family enters the closure.
- **No unit test is added, and none is owed**: this change adds no behaviour. What it changes is what the runner reports about the tests that already exist, and the evidence for it is the suite running green with the collector enabled, which every gate and the `Frontend` CI job now do.

## Checklist

- [x] `bash scripts/verify-full.sh` passes on this branch, rebased onto current `main`.
- [x] Behavior changes are covered by unit tests, and the coverage gate passes. — no behaviour changed; the service's coverage gate is untouched and still enforces 85%.
- [x] Affected documentation is updated in this change set.
- [x] `bash scripts/review-obligations.sh` was run, and every row it reported is answered — by a test, by a page, or by why nothing is owed there.
- [x] `CHANGELOG.md` is untouched — it is written by the release pull request alone.
- [x] Every new C# file carries the repository's standard header, applied by `scripts/verify-fast.sh` rather than typed, and no file gained a second copyright line, an author tag, a name, or a contact detail. — no new file of any kind; the one source file changed already carried its header.
- [x] Nothing here is under terms MailFathom could not distribute under Apache-2.0. A new dependency, service, container image, or externally sourced sample has its row in `THIRD_PARTY_LICENSES.md` in this change set. — `@vitest/coverage-v8` is MIT per its own npm metadata, recorded in the client's test-only row, and it reaches no artifact.
- [x] No credential, token, private key, real mailbox data, or personal information is in the diff.


docs/operations/agent-workflow.md
Krzysztof318, line 1: Ci workflow?

Krzysztof318: `CI` already collects and prints it, and that is why `.github/workflows/` is not in this diff.

The `Frontend / Check, build, and test the client` job's `Run both packages' unit suites` step is `pnpm --dir frontend run test`, which is the same command both verification gates run. Coverage is enabled in `frontend/vitest.config.ts` rather than passed as a flag, so that step now prints the summary in its own log with no change to the workflow file. Nothing there had to be told about it, which was the point of enabling the provider in the configuration instead of adding a `test:coverage` script beside `pnpm test`.

What `CI` does **not** do is upload the HTML report as an artifact, the way `build-test-format-and-migrations.yml` uploads the service's under `Upload code coverage`. That is deliberate and is listed under *Out of scope* on #1441: the summary in the job log is what this change delivers, and retaining the report as a downloadable artifact is a separate decision about what is worth keeping from a run nothing gates on.

Happy to add the upload here if you would rather have it — it is one `actions/upload-artifact` step pointed at `artifacts/coverage/client/`, and it would need a line in `docs/operations/local-development.md` § *The client's suite* saying so. Say the word and I will; otherwise this stays as it is and the out-of-scope note stands.

Issue #1477, opened 2026-09-02
Decide whether the client may show a message's original HTML, which ADR 0024 refuses


Krzysztof318:
## Context

The design project's message head now carries a `code` control, titled *show the full HTML version*, which opens the message's own markup in a pane of its own. [ADR 0024](https://github.com/Krzysztof318/MailFathom/blob/main/docs/decisions/0024-rendering-mail-in-the-client-as-a-closed-document-tree.md) refuses exactly that, and refuses it as the load-bearing property rather than as a preference: `ClientMailBodyEndpoint` sets `IncludeMailDocument` on every read and `IncludeSanitizedHtml` on none, so the client is never handed mail HTML and therefore renders none, and a lint rule makes `dangerouslySetInnerHTML` and its four siblings unwritable under `frontend/src/`. That record's own table names the shape of the contract as what holds *no script, embedded object, form, or event handler executes*, with nothing about the browser relied on.

So this is not a control to build. It is the question ADR 0024 answered, asked again with a use that record did not weigh: a reader who wants to see what the sender actually sent, because the reduction lost a document they need.

Two things make it more than one button. Serving the markup at all changes the client body route — the service already produces a sanitized representation for `IncludeSanitizedHtml`, and asking for it is ADR 0024's option A2, which that record considered and did not take. And where the markup is rendered decides whether the pane stops being ordinary elements in the application's own document, which is the same record's option B. Its remote-picture clause holds either way and is the sharpest edge here: every remote address is removed while the document tree is built, so *show the original* is precisely the path on which a read receipt could come back.

ADR 0024 is still `proposed`, so the answer either amends it or supersedes it, and `docs/AGENTS.md` decides which. This issue produces that record and nothing else. `CLAUDE.local.md` states the other half: where implementation proves a design wrong, the design project is corrected to the shape that actually ships, so a *no* here is written there as well as in the tree.

## User stories

**As somebody reading a message that did not reduce well** I want to see what the sender actually sent, **so that** a document the reduction lost is not lost to me.

**As the person who has to trust this client** I want to know whether the safety property ADR 0024 states still holds, **so that** a convenience did not quietly cost me the guarantee I was given.

## Acceptance

- An ADR settles whether the client may show a message's original HTML, and it amends or supersedes ADR 0024 rather than standing beside it.
[...]

PR #1487, opened 2026-09-02
[#1486] Count the frame's browsing context among what the full-HTML surface rests on the browser for

### LINKED ISSUE #1486: ADR 0024 undercounts what the full-HTML surface rests on the browser for: three properties, not two
## Context

ADR 0024's second table names what holds each of the five safety properties on the full-HTML surface, and the prose under it says how many of them now rest on the browser. The count is wrong by one, in the direction that understates it.

The table attributes a browser mechanism to three rows: row one to the sandboxing flags, row five to those flags together with the scheme allow-list, and row three — *message style cannot escape* — to the frame being its own browsing context. The prose says two. Row three's isolation is as much the engine's guarantee as the sandbox attribute rows are; it reads as free because a frame feels structural rather than like a capability, which is exactly how it got missed.

That count is not decoration. It is what a later proposal in this space is judged against — a third surface, a change to how the frame is composed, a head whose engine behaves differently — and the record's own driver says the safety property has to be statable rather than merely present. A reader taking the count at face value would believe one more property is off the platform than actually is.

The defect entered on #1477 and merged in #1480. It was raised as a P3 on that pull request's final review, and the fix was written but not pushed before the merge landed.

## User stories

**As somebody weighing a change to the full-HTML surface** I want the record's count of what rests on the browser to match the table above it, **so that** I am judging the proposal against what is actually true rather than against an undercount.

## Acceptance

- The prose under the second table says three of the five properties rest on the browser, and names which three.
- The three places that state the count agree: the paragraph under the table, the *Consequences* bullet, and the *Revisit if* clause.
- The table row for *message style cannot escape* says which mechanism holds it and that it is counted with the sandbox rows, so the count can be checked against the table without inferring anything.
- The two properties that stay off the platform are named as such: one held by a build that fails, one by a removal the service already performed.
- Nothing else in the record changes. This corrects a count, not a decision.

## Out of scope

- Reopening any part of what #1477 settled.
- #1484 and #1485, which build the two halves of the surface.

## Dependencies

None.

Krzysztof318:
## What this corrects

ADR 0024's second table names what holds each of the five safety properties on the full-HTML surface, and the prose under it counts how many now rest on the browser. It said **two**. The table attributes a browser mechanism to **three** rows:

| Row | What holds it | Browser? |
|---|---|---|
| 1 — nothing executes | the sandboxing flags | yes |
| 2 — no value becomes markup | a build that fails | no |
| 3 — style cannot escape | the frame's own browsing context | **yes — the one that was missed** |
| 4 — nothing is fetched unasked | a removal the service already performed | no |
| 5 — a link cannot navigate the application | the flags and the scheme allow-list | yes |

Row three's isolation is as much the engine's guarantee as the sandbox attribute rows are. It read as free because a frame feels structural rather than like a capability, which is exactly how it got missed.

The count is not decoration: it is what a later proposal in this space gets judged against, and the record's own driver says the safety property has to be *statable*, not merely present. Corrected in all three places it was stated — the paragraph under the table, the *Consequences* bullet, and the *Revisit if* clause — and the table row now says which mechanism holds it and why it is counted there, so the count can be checked against the table without inferring anything.

Nothing else changes. This corrects a count, not a decision.

## Why it is a separate pull request

The defect entered on #1477 and merged in #1480. It was raised as a P3 on that pull request's final review, and this fix was written but not pushed before the merge landed — so it arrives on its own issue rather than as a late commit on a closed branch.

## Breaking changes

None. Documentation only.

## Verification

`scripts/verify-full.sh` — 287 passed, 0 failed. `scripts/review-obligations.sh` reports no tests, pages, or registers owed.

Closes #1486

https://claude.ai/code/session_01BT7ZES6jyEX5k5ommU1rYm


Krzysztof318:
**P2 — the fourth "two of the five", in the B4 options entry — fixed in `1d1b79cd7`.**

Correct, and it is the same defect this pull request exists to remove rather than a near miss: a record contradicting itself about a number is no better when three of the four statements agree. I corrected the places the count is *argued* — the paragraph under the table, the Consequences bullet, the Revisit-if clause — and did not check the place it is merely *restated*, which is exactly where a reader comparing the options section against the table would have landed.

The entry now reads "three of the five properties onto the browser — two onto the sandboxing flags and one onto the frame's own browsing context", naming the mechanisms the way the other three do, so the four statements read as one fact rather than as four sentences that happen to agree.

Verified by grep afterwards that no statement of the old count survives anywhere in the record: four sites say three, and the Revisit-if clause names both mechanisms whose divergence would cost all three at once.

`scripts/verify-full.sh` — 287 passed, 0 failed.

Issue #1618, opened 2026-09-03
Answer a swipe on a message row — left to reply, right to archive — and a long press for selection


Krzysztof318:
## Context

The design's prototype answers a touch on a message row with more than a tap, and the behaviour is
specific enough to be implemented rather than interpreted:

- **A swipe left opens the conversation and starts a reply.**
- **A swipe right archives the message**, and the archive is remembered so it can be undone — which is
  exactly the toast with a single *undo* action stage 3 delivered (#1604).
- **The threshold is 96 CSS pixels of horizontal travel.** Below it the row springs back and nothing
  happens.
- **Vertical movement cancels the gesture.** More than 26 pixels of vertical travel, exceeding the
  horizontal, abandons it — a person scrolling a list is scrolling, not swiping.
- **The tap that would otherwise follow a completed swipe is suppressed** for a moment afterwards, so
  an archive does not also open what it just archived.
- **A long press opens the row's context menu**, at 420 ms and with a short vibration where the
  platform offers one, cancelled by any movement of the finger and suppressing the tap that would
  otherwise follow it. That menu is what carries the multiple-selection actions a desktop reaches from
  the toolbar — its first item starts the selection — so the press opens a menu rather than a
  selection. The menu itself is #1633; what belongs here is that this gesture and the swipe share one
  finger and must not fire together.

None of this exists in `frontend/src`, and the message list is drawn for a pointer that hovers.

**What a gesture may do is bounded by the stage that owns the act.** Archiving is a mailbox write, and
every mailbox write belongs to stage 5 (#1207) — so where this lands before that stage does, the swipe
performs whatever the client can already perform and the rest arrives with the writes. What this issue
owns either way is the gesture, its thresholds, its cancellation, and its reporting.

## Scope

- The horizontal swipe on a message row, both directions, with the threshold, the cancellation rule,
  the spring-back, and the click suppression the design draws.
- The affordance while the gesture is in progress: what the row shows under the finger as it moves,
  per the design.
- The long press only as far as it meets the swipe: the two gestures begin on the same touch, so the
  press is armed and cancelled by the rules this issue defines and hands off to the menu #1633 builds
  rather than opening one of its own.
- The reporting: a completed swipe reports through the toast surface (#1604), and where the act has a
  way back, the undo is that toast's single action.
- Accessibility, which a gesture cannot be the only route to: everything reachable by swipe or long
  press is reachable another way — a control, a menu, or the keyboard — and nothing is gesture-only.
[...]

Issue #1617, opened 2026-09-03
Take every screen the client already has through the four compositions the design draws


Krzysztof318:
## Context

The shell composes itself for four widths in the sibling issue. This one takes every screen the client
already has through the same four, because a bottom bar around a screen still drawn for a mouse is
half a phone client.

The screens that exist today: sign-in, the mailbox tree, the message list, the conversation, the
reading pane and its attachment strip, the full-HTML surface, search and its filters, settings, and
the notification centre and toast surface stage 3 delivered. Each of them is drawn in the design
project at the desktop, tablet, fold and phone compositions, and what the design changes between them
is more than a width: avatars and icons take different sizes, toolbar controls drop their labels and
become icons at the tablet width already, rows grow to a height a finger can hit, and panels that sit
beside content on a desktop are drawers.

`frontend/src/AGENTS.md` already states the bar this meets — every screen usable from 320 CSS pixels
upward with no horizontal scrolling, nothing hidden by width alone, `pointer: coarse` asked where a
target must fit a finger, `hover` asked where an affordance would otherwise exist only under one — and
this is the issue where the existing screens are held to it rather than assumed to meet it.

## Scope

- Every screen listed above, at the four compositions the design draws, including its empty, loading,
  and error states, which are the ones a narrow width breaks first.
- Touch sizing: row heights, control sizes, and the toolbar behaviour the design gives the tablet and
  phone.
- Anything that exists only under `hover` today, given a route a touch device can reach.
- Text that has to reflow rather than truncate, and anything that currently assumes a wide pane.

## Acceptance

- [ ] Every screen listed in the context composes for the desktop, tablet, fold and phone as the
      design draws it, including its empty, loading and error states.
- [ ] Every screen is usable from 320 CSS pixels upward with no horizontal scrolling of the page, and
      nothing is hidden by width alone without being reachable somewhere else.
- [ ] Touch targets, row heights, and toolbar controls take the sizes the design gives them at each
      composition.
- [ ] Nothing is reachable only under `hover`; every such affordance has a route a touch device has.
- [ ] No component, hook, or screen branches on the head; the compositions come from width and
      `pointer: coarse`.
- [ ] Unit tests beside the sources they cover, driven by viewport, for each screen whose composition
      genuinely differs rather than one blanket test.
- [ ] The browser suite covers the phone composition of the mail screens.
- [ ] Every screen is held against the design project as two screenshots per composition that differs,
      and every difference is named.

## Out of scope

[...]

Issue #1614, opened 2026-09-03
Make the client build and run as an Android head, from the same bundle the other two serve


Krzysztof318:
## Context

`frontend/src-tauri/` builds a desktop shell and nothing else: `tauri.conf.json` carries no Android
configuration, the crate declares no mobile entry point, `capabilities/` holds one file written for a
desktop window, and no Android target is installed for any toolchain here. ADR 0021's proof of concept
produced an APK from this tree, so the work is configuration and an entry point rather than a port —
but none of it is committed, and a build today does not exist to be run.

What this child delivers is a shell that builds and runs on a device or an emulator, from the same
bundle the other two heads serve, so that the CI job beside it has something to build and the
composition work beside it has somewhere to be tried.

The parts, from Tauri's own prerequisites and CLI: the four Android Rust targets, the Android SDK, NDK
and JDK the build reads through `ANDROID_HOME`, `NDK_HOME` and `JAVA_HOME`, `tauri android init` to
generate the Gradle project, a mobile entry point in the crate beside the desktop one, and the
application identifier, the minimum SDK level, the icons, and the one permission the client actually
needs, which is network access.

Two decisions belong to this issue rather than to the ADR beside it, because they are about this
repository's own layout: **whether the generated Gradle project is committed or regenerated on every
build** — regenerating keeps one source of truth and costs a step in CI, committing keeps the build
reproducible and puts a generated tree under review — and **what the application identifier and the
minimum SDK level are**, which are stated once and then never guessed at again. Both are recorded in
the issue's own resolution and in `frontend/AGENTS.md` where the client's toolchain is described.

## Scope

- The Android target in `tauri.conf.json` and in the crate: the mobile entry point, the identifier,
  the minimum SDK, the icons, and the permissions, which are the ones the client needs and no others.
- The Rust targets and the toolchain prerequisites, written down where a person setting this clone up
  will find them — `docs/operations/local-development.md` for the environment, `frontend/AGENTS.md`
  for what the client stack expects.
- `capabilities/` for the Android head: the same operations the desktop window is permitted, granted
  explicitly rather than by a plugin's default set.
- Whatever the desktop head's `run-tauri.ts` wrapper has to learn so that the version comes from
  `Version.props` on this head too, rather than a second number appearing in a Gradle file.
- Proof it runs: the client reaches a deployment, signs in, and draws mail on an emulator and on a
  device.

## Acceptance

[...]

PR #1677, opened 2026-09-05
[#1652] Take the Android system bars off the page, and write down how the head reaches a deployment

### LINKED ISSUE #1652: Verify the Android head signs in and draws mail, on an emulator and on a physical device
## Context

#1614 made the client build and run as an Android head: `pnpm android:build` produces an APK, the
generated Gradle project is committed and reviewed like source, the version comes from `Version.props`,
and the head's capabilities are declared. What that change could prove locally stops one step short of
its own acceptance. The APK installs, launches, and draws the sign-in screen on an emulator, and a
request really leaves the device — a TLS handshake reached a listener standing in for a deployment, and
the client reported the failure it should when nothing answered. What was never exercised is the part
after that: a real deployment, a real credential, a mailbox, and a screen with mail on it. No deployment
and no mailbox exist on the machine that built it, and no physical device was attached.

That leaves one acceptance item of #1614 unmet rather than reinterpreted, which is why it is carried
here instead of being closed with the rest.

## Scope

- Sign in from the Android head against a running deployment over `/api/client`, on an emulator, and
  read mail on the screen the other two heads draw.
- Do the same on at least one physical Android device, which is the case the emulator cannot answer:
  a real network path, a real WebView build, and the vendor's own window insets.
- Describe what was observed rather than capturing it. A signed-in client shows real mail, so a
  screenshot of one is personal data and stays out of the issue, the pull request, and the commit.
- Correct in `frontend/` whatever the run finds, or open the issue that owns it where the defect
  belongs to a child that already exists — the credential store is #1619, the system notification
  #1620, and the back gesture #1616.

## Acceptance

- [ ] The client signs in on an emulator against a running deployment and draws mail, and what was
      observed is described in this issue.
- [ ] The same is done on at least one physical device, and the device, its Android release, and what
      differed from the emulator are named.
- [ ] Anything the run found is fixed here or filed against the child that owns it, and this issue
      names which.
- [ ] `docs/operations/local-development.md` says how the head is pointed at a deployment for this
      run, if what was done is not already what that page describes.

## Out of scope

- The credential store (#1619), the system notification (#1620), and the back gesture (#1616), each of
  which is its own child. A defect this run finds in one of them is reported there rather than fixed
  here.
- Any composition work. Whether a screen looks right on a phone is #1617 and #1618; this asks only
  whether it signs in and shows mail.
- The nightly job (#1615). This is a run somebody performs, not a check a pipeline gains.

## Dependencies

Waits on #1614, which is the head this is run against.

Krzysztof318:
The Android head was run against a real deployment on an emulator — signed in over `/api/client`, drew a mailbox, and
opened a message. What that run found is fixed here, and how the run was set up is written down so the next one is not
reconstructed from scratch. The full account of what was observed is in the issue.

## The defect the run found

`MainActivity` calls `enableEdgeToEdge()`, which makes the window full-bleed, and nothing then took the insets off the
content. So the WebView spanned the whole display and the client drew under the system bars: the mail space's top bar
sat beneath the clock and the bottom navigation beneath the gesture handle.

The client is not what is wrong. It already asks for this the way the web platform offers it — `viewport-fit=cover` on
the document and `env(safe-area-inset-*)` behind the `--spacing-safe-*` tokens in `styles.css`, which nine components
consume. Android's WebView is what does not answer: it reports those insets for a display cutout alone and never for
the system bars, so all four measured `0px` on an emulator that has both.

Fixed in the shell rather than in the page, so nothing in `frontend/src/` learns that a third head exists.
`MainActivity` now pads the content view by the system bars, a display cutout and the keyboard, and consumes them.
Measured on the emulator before and after: the viewport went from 699 to 651 CSS pixels — the 24 dp status bar and the
24 dp gesture bar are outside it — and the screens sit where they should.

What padding costs is named in the comment beside it: the strip behind each bar is painted by the window background
rather than by the page, so a head whose theme is set against the system's shows the system's colour there. Lifting
that means handing the measured values to the page as the `--spacing-safe-*` tokens instead of padding at all, which
needs a WebView reference and a first delivery that cannot race the load.

## Documentation

`docs/operations/local-development.md` gains *Pointing the Android head at a deployment*, which the run needed and the
page did not have: the `10.0.2.2` alias and the two answers a physical device has instead, the two separate clear-text
permissions and where each is granted, `client.conf` and what `adb install -r` and `pm clear` do to it, and the
debuggable WebView socket that makes a run observable without a screenshot of anybody's mail. `frontend/README.md`
gains the inset decision in the table of what the committed Gradle project decides.

`THIRD_PARTY_LICENSES.md`'s Android row said no Kotlin of ours exists to choose a dependency for. It does now — this
activity reaches `androidx.core` under `appcompat` — so the clause says what is true instead. No declaration moved and
no census changed: `androidx.core` was already inside the locked 80-artifact closure, and the build verified against
`gradle.lockfile`.

## What is not done

**The physical device.** No Android device is attached to this machine and attaching one is not something a session
can do, so the third acceptance item and the second half of the second one stay open — the issue names what is left
for whoever runs it. The inset fix above was measured against the emulator's 24 dp bars; a device with a cutout, a
punch-hole, or a taller gesture inset is exactly the case that would show it being read from the platform rather than
assumed.

**Two things for the owner to decide, neither touched here.**

*ADR 0027 carries a claim this run contradicts.* Its paragraph on the safe-area seam says an Android device with a
gesture bar and a cutout "is where those values stop being zero". The gesture-bar half is false — the WebView reports
`0px` for every system bar — and after this change the cutout half is false on Android too, since the page no longer
overlaps one. An accepted ADR is closed and `docs/decisions/` is a protected path, so nothing here edits it; whether
that is worth a superseding record is yours.

*A composition observation, deliberately not acted on.* On the phone the mail space draws the selection bar above the
reading pane once a message is open, because opening a row makes it the sole selection and that bar replaces the
toolbar at every width. Whether that is what the phone screen should draw is composition, which this issue puts on
\#1617 and #1618, so it is recorded rather than changed — and it is worth holding against the design project before
anything is done to it.

## Verification

`scripts/verify-full.sh` passed, 288 of 288. The Android head is gated by neither verification script nor any
pull-request check, so what proves this one is `pnpm android:build` run by hand — done twice here, and the second APK
is what the measurements above were taken on.

Closes #1652


Krzysztof318:
On the one candidate the review dropped as uncheckable from this repository — whether `app.path().app_config_dir()` on Android resolves where the new page says `client.conf` lives: it does, and the run is where that came from rather than a reading of Tauri. The emulator carried a file left by the #1614 run at exactly `/data/data/io.github.krzysztof318.mailfathom/client.conf`, read back with `adb shell run-as`, holding `service-address` and `permit-clear-text`; the sign-in screen drew that address as one that could not be changed, and clearing the file was what let the head reach a deployment at all. So the path in the page is an observation rather than an inference. Dropping it was still right — nothing in the tree proves it.

Issue #1715, opened 2026-09-06
The client's browser suite is red on main, and has been since the design-parity change merged


Krzysztof318:
## Context

Three checks in the client's browser suite fail on `main`, and have since #1710 merged: that pull request's
`Frontend / Check, build, and test the client` job is recorded as `FAILURE` and the change was merged over it, so the
required check has been red on every client change since. Reproduced locally against `origin/main` with an unmodified
`frontend/tests/client.spec.ts`, twice, with identical results — these are deterministic failures rather than the
browser flakes `frontend/tests/AGENTS.md` warns about.

The three, with what each reports:

- `runs nothing the markup carries, and reaches no host but its own until the reader asks` —
  `expect(getByText(/permits no script at all/)).toBeVisible()` finds no element. Either the sentence the markup
  surface draws about the frame changed, or the surface no longer draws it.
- `draws the region again when it is retried, and hands the keyboard into it` — a strict-mode violation:
  `getByRole('button', { name: 'Reply — not built yet' })` now resolves to two elements, one in the
  `Mail actions` toolbar and one in the `What is open` region. The reading pane grew a second control carrying the
  same accessible name, so the query naming one is now ambiguous.
- `draws the mail screens at the phone composition, with its own row height and nothing over the question` — a 30s
  timeout reading the bounding box of `getByRole('button', { name: /^New message/u })` at the phone frame. The control
  the check measures is either absent or not laid out at 390 × 844.

The first two say the client changed and the assertions did not follow; the third says either the same or that the
phone composition lost a control. Which of the two each is, is what this issue settles — the suite is the record of
what a browser can say about the client, and a red one that everybody merges over stops being that.

## User stories

**As** the owner **I want** the browser suite green on `main` **so that** the required check means something on the
next client pull request instead of being a red square everybody has learned to merge past.

**As** an agent taking a client issue **I want** to know which failures are mine **so that** a pull request's red
check is a finding about the change rather than something inherited from the branch it started on.

## Acceptance

- The three checks named above pass against `main`, and `pnpm test:browser` is green with no test skipped, no retry,
  and no assertion loosened to something that would pass either way.
- Each is resolved by deciding which side was wrong and fixing that side: where the client regressed, the client
  changes; where the client is right and the assertion is stale, the assertion changes and the reason is in the diff.
  A second control carrying an existing accessible name is settled as an accessibility question first — two controls a
  screen reader announces identically is a defect in the screen before it is a defect in a locator.
[...]

PR #1717, opened 2026-09-06
[#1700] Give the client one corpus of example mail, imported by every suite that needs it

### LINKED ISSUE #1700: Give the client one corpus of example mail, imported by every suite that needs it
## Context

Every client check that needs mail invents it again. `frontend/tests/client.spec.ts` declares a session answer, accounts, folders, `emails*`, `messages/*` and `messages/*/body*` inline; a throwaway parity spec under `frontend/.playwright/` declares the same shapes a second time and adds `threads/*`, `search`, `drafts`, `notifications`, the mutation routes and `signals`; and a session that wants to look at a screen writes a third copy before it can take a screenshot. Three copies of one corpus is three places to be wrong, and the shapes are not obvious: a body fixture with `plainText: null` is refused and the reading pane draws *could not be read: unreadable*, which reads as a client defect rather than as a malformed fixture.

The cost is paid on every client task, and it is paid before any work starts. It is also what makes the design-parity loop slow: reaching a screen with data is most of the effort, and it is effort spent reproducing something that already existed in another file.

Nothing governs this beyond `frontend/tests/AGENTS.md` § *What is faked, and where*, which decides that the network boundary is `MailFathomTransport` and that nothing patches `fetch`, starts a server, or adds an HTTP mocking package. This body is the governing text for the rest.

## User stories

**As** an agent taking a client issue **I want** one corpus of example mail already written, **so that** reaching a populated screen is an import rather than an hour of guessing response shapes.

**As** the owner **I want** every client suite to read the same example data, **so that** a shape that drifts from the service is wrong in one place rather than in three that disagree.

## Acceptance

- A fixture module under `frontend/tests/fixtures/` holds one corpus: the session answer, accounts, folders, the mail timeline, messages and their bodies, threads, search results, drafts, notifications, and the mutation and signal answers the client reads.
- The corpus is data rather than routing. It exports values, and each consumer decides how they reach the client — `page.route` in the browser suite, a transport function in a unit test.
- `frontend/tests/client.spec.ts` imports it and declares no fixture of its own, so the committed browser suite is the first proof that the corpus is sufficient.
- The corpus covers the states a screen has to be looked at in, not only the resting one: an empty folder, a message with no plain text alternative, a thread long enough to collapse, a failed synchronization, and an account behind.
- Every message body carries a `plainText` object, and the module's own documentation says why a `null` there is refused.
- The mail in it is invented and reads as ordinary correspondence, with no address, name, or subject taken from anywhere real.
- The client's unit suite and browser suite both pass against it unchanged in what they assert.

## Out of scope

- Binding the corpus to the recorded HTTP contract. That is its own issue and it waits on this one.
- Serving the corpus to a running client. The transport that does it is its own issue.
- Any change to what a suite asserts. This moves where the data comes from and nothing else.

## Dependencies

None.

Krzysztof318:
Closes #1700

The browser suite carried its own mail inline. A session answer, one account, two folders, a
generated timeline, one message and its body were all declared in `client.spec.ts` beside the checks
that read them — fine for one suite and wrong for two, because the second one to need a populated
screen invents a mailbox of its own and the two drift apart with nothing to notice.

`frontend/tests/fixtures/` is now one corpus, and it is **data rather than routing**: every module
exports values and none of them knows what a `page.route` or a transport function is, so a unit test
reaches it exactly as the browser suite does. Each consumer decides how the values get to the client.

## What the corpus holds

Split by what each module is about rather than by which check reads it:

- `deployment.ts` — the deployment and the person: the credential and the `Authorization` header it
  produces, the session answer, the display name, the client preferences, the accounts and the
  folder tree. The resting tree is byte-identical to what the spec declared before, so nothing it
  asserts moved.
- `messages.ts` — one message drawn every way the reader draws it: the document tree, the sender's
  own markup, remote images asked for and withheld, an attachment, and the message that carries no
  plain text part at all.
- `mail.ts` — the mailbox as rows. One shape covers the timeline, a search result and a
  conversation's messages, because the service publishes a search result as a list row plus two
  fields and a conversation's `email` as that same row field for field. `timelinePage(from)` answers
  a cursor rather than a fixed page, so paging forward and backward is the corpus's arithmetic rather
  than the spec's.
- `drafts.ts`, `notifications.ts`, `changes.ts` — a draft and an answer to one, a notification page
  covering all three target shapes and both read states, and both directions a change travels: the
  mutation answers, the record states, and one payload of each of the five signal kinds.

It covers the states a screen has to be looked at in and the resting one does not show: an **empty
folder**, a **message with no plain text alternative**, a **thread long enough to collapse** (five
messages, four of them behind the fold `Thread.tsx` draws), a **synchronization that failed**, and an
**account behind**.

Every message body carries a `plainText` object, and `messages.ts` says why rather than leaving the
next reader to decide: `mailBody.ts` refuses a `null` there and the reader is told the message could
not be read, so a fixture with a `null` in it would be exercising the failure path while looking like
an ordinary message. The `plainText` on the *message* route is a different thing — a `boolean` saying
whether the message carried a text part — and the module says that too, because the two names are one
transposition apart.

Nothing in it is anybody's. Every address is under a reserved name (`.invalid`, `.example`), and
every person, subject and sentence is invented.

## What is deliberately not here

- No unit test imports it yet. The two Vitest projects are rooted at `src/Client.Backend` and
  `src/Client.App`, so an import of `../../../../tests/fixtures/…` would cross the package boundary
  that layout exists to refuse; the issue's own scope is the corpus, not its consumers.
- No module imports a type from `@mailfathom/client-backend`. The corpus states what the service
  answers with, which is what a fake is for — typing it against what the client parses would make a
  mistake in the parser invisible.
- The draft listing is a bare array rather than an envelope: nothing in `Client.Backend` reads that
  route yet, and inventing an envelope would be exactly the drift this directory exists to end.
- `sessionAnswer.version` is `'0.0.0'` and the consumer substitutes it. Only a build knows the real
  one, and `AGENTS.md` allows nobody to retype it.

## The browser suite is red, and not from this change

`Frontend / Check, build, and test the client` will fail on three checks that fail on `main` as well:
the markup-frame sentence, a `Reply — not built yet` name that now resolves to two controls, and the
phone composition's `New message`. They were reproduced against `origin/main` with an unmodified
`client.spec.ts`, twice, with identical results; #1710's own client job is recorded as `FAILURE` and
was merged over, so the required check has been red on every client change since. They are #1715 and
are not this pull request's to fix.

## Verification

- `scripts/verify-fast.sh` and `scripts/verify-full.sh` pass over the pushed tree; the client unit
  suite is 2916 tests and unchanged in what it asserts.
- `pnpm test:browser` gives the same 45 passed / 3 failed against this branch as against
  `origin/main` with the spec reverted and the corpus removed — the three are the #1715 failures and
  nothing else moved.
- `$check-docs-licenses`: docs `pass`, changelog `n/a`, licenses `n/a` — no dependency is added,
  upgraded or newly used, so no register row, no census and no bundled notice moves.

### DIFF
--- a/docs/architecture/solution-structure.md
+++ b/docs/architecture/solution-structure.md
@@ -184,7 +184,10 @@ real browser, which is what answers the questions jsdom structurally cannot — 
 a document with a history, and the requests a page actually issued; it belongs to neither package, imports neither, and
 therefore does live under `frontend/tests/`, beside the contract governing both.
 [`frontend/tests/AGENTS.md`](https://github.com/Krzysztof318/MailFathom/blob/main/frontend/tests/AGENTS.md) is that
-contract.
+contract. `frontend/tests/fixtures/` beside them is the third thing in that directory and belongs to neither package
+either: one corpus of invented example mail, exported as values, which whatever needs a populated screen imports rather
+than writing out again — the browser suite reaches it with Playwright's own routing, and a unit test would hand it to a
+transport function.
 
 What the workspace builds is a directory of static files under `frontend/src/Client.App/dist/` and nothing else, so no
 Node process joins any deployment shape.
--- a/frontend/README.md
+++ b/frontend/README.md
@@ -332,7 +332,13 @@ produces, and the requests the page actually issued. It needs a browser of its o
 `pnpm exec playwright install chromium` — which is why neither verification gate runs it and the pipeline does, on every
 pull request that reaches this stack. Its configuration is `playwright.config.ts` and its specs are under `tests/`.
 
-[`tests/AGENTS.md`](tests/AGENTS.md) is where both suites' policy is decided, including which check belongs to which.
+`tests/fixtures/` beside them is one corpus of example mail — the session, the accounts and folders, the mail, the
+conversations, the drafts, the notifications, and what a change answers with — imported by whatever needs a populated
+screen rather than written out again per check. It is data and no consumer of it is assumed: the browser suite reaches
+it with `page.route`, and a unit test would hand it to a transport function.
+
+[`tests/AGENTS.md`](tests/AGENTS.md) is where both suites' policy is decided, including which check belongs to which,
+and where the corpus's own rules are.
 
 ## Whitespace is decided in `.editorconfig`
 
--- a/frontend/tests/AGENTS.md
+++ b/frontend/tests/AGENTS.md
@@ -19,9 +19,10 @@ question is answered again rather than reworded.
   could reach one only through a relative path out of the package, which is the one thing that boundary exists to
   refuse. A test inside the package inherits it instead: a `Client.Backend` test cannot import React, exactly as its
   source cannot, and nothing has to check that it did not.
-- `frontend/tests/` therefore holds this file and the browser suite beside it, and nothing else. A unit test written
-  here would resolve neither package, which is the whole of the argument above; the browser suite is what can live here
-  precisely because it imports neither — it drives a built bundle over HTTP rather than importing a module out of one.
+- `frontend/tests/` therefore holds this file, the browser suite beside it, and [the corpus](#the-corpus) both suites
+  read — and nothing else. A unit test written here would resolve neither package, which is the whole of the argument
+  above; the browser suite is what can live here precisely because it imports neither — it drives a built bundle over
+  HTTP rather than importing a module out of one, and the corpus is data rather than a test.
 - **The two suites are told apart by the name.** A unit test is `*.test.ts` or `*.test.tsx` beside its source; a browser
   spec is `*.spec.ts` under this directory. Each runner's default finds its own and neither finds the other's, so a file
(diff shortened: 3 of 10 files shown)


Krzysztof318:
The `Frontend / Check, build, and test the client` job on this head reports **3 failed, 45 passed**, and the three are exactly the ones #1715 records — the markup-frame sentence at `client.spec.ts:992`, the ambiguous `Reply — not built yet` name at `:1080`, and the phone composition's `New message` bounding box at `:1161`. No other check in the client flow failed: lint, the type check, the unit suite and the build all passed, and `Required CI` is red only because that job is.

That is the same set the suite gives against `origin/main` with this branch's spec reverted and the corpus removed, so the pipeline now confirms in CI what was measured locally: the failures predate this change and none of them reads a fixture this pull request adds.

PR #1721, opened 2026-09-07
[#1703] Hold the client's fixture corpus against the recorded HTTP contract

### LINKED ISSUE #1703: Check the client's fixtures against the recorded HTTP contract, where neither stack's filter would run it
## Context

Nothing holds the client's idea of the service against the service's own. The types in `Client.Backend` are written by hand from the documentation, and the fixtures every client suite answers with are written by hand too, so both ends can agree with each other and disagree with the service while every gate reports green. `backend/tests/PublicSurfaces.UnitTests/http-api-contract.json` already records the whole HTTP surface byte for byte and nothing under `frontend/` reads it.

The failure this admits is not hypothetical and not loud. An enumeration is camel-cased over the MCP surface and Pascal-cased over `/api/client`; a fixture that copies the wrong spelling passes the linter, the type check, both unit projects and the browser suite, and is discovered on somebody's deployment.

Where the check lives is the part worth deciding once. It cannot sit in either stack: `scripts/resolve-changed-stacks.sh` and `ci.yml`'s `detect-changes` name no backend path in the client filter and no client path in the server's, so a check written on either side simply would not run when the other side moved. `scripts/test-agent-workflow.sh`, run by the `Repository contracts` workflow, is the existing home for an invariant that belongs to neither stack and has to hold across both.

## User stories

**As** the owner **I want** the client's example data checked against the recorded HTTP contract, **so that** a wrong field name or a wrong casing fails in the pipeline rather than on a deployment.

**As** an agent adding a route to the client **I want** the mismatch named, **so that** I correct a spelling instead of debugging a screen that renders nothing.

## Acceptance

- `scripts/test-agent-workflow.sh` compares the client's fixture corpus against `http-api-contract.json` and fails naming the route, the field, and both spellings when they disagree.
- Field names and enumeration spellings are both covered, because the casing difference between the two surfaces is the failure this exists for.
- A route the contract records and the corpus does not cover is reported rather than failed, so the check states the gap without blocking work that has not reached that route yet.
- A fixture shape the contract does not know at all fails, because that is a route the client invented.
- The check's own reasoning is written where it lives: why it is not in either stack, and what would break if somebody moved it into one.
- It runs on every pull request through the existing `Repository contracts` workflow, with no second workflow added.

## Out of scope

- Generating the client's types from the contract. That is a larger decision about how the two stacks are coupled and is not taken here.
- Any change to `http-api-contract.json` or to how it is recorded. It is the reference, and this reads it.
- The MCP surface's own contract. The client does not speak it.

## Dependencies

Waits on #1700: there has to be one corpus to check before checking it means anything, and a check written against three scattered copies would be three checks.

Krzysztof318:
Closes #1703

Nothing held the client's idea of the service against the service's own. The corpus under
`frontend/tests/fixtures/` is written by hand from the documentation, and so are the types in
`Client.Backend` that read it — so both ends agreed with each other and were free to disagree with the
deployment while the linter, the type check, both unit projects and the browser suite all reported green.
`backend/tests/PublicSurfaces.UnitTests/http-api-contract.json` already records the whole HTTP surface byte
for byte, and nothing under `frontend/` read it.

## Where the check lives, and why it cannot live anywhere else

`the_client_corpus_states_what_the_contract_records` is a contract of `scripts/test-agent-workflow.sh`,
which the `Repository contracts` workflow already runs on every pull request. No second workflow is added.

It belongs to neither stack because neither stack's change filter would run it when the other side moved.
`scripts/resolve-changed-stacks.sh` and `ci.yml`'s `detect-changes` name no `backend/` path in the client
filter and no `frontend/` path in the service's — the disjointness
`the_stacks_change_filters_name_no_path_in_each_other` already asserts. So a field renamed under
`backend/src/Host/Api/` reaches this repository as a moved line in `http-api-contract.json`, touches no
client path, and the client job reports green on a corpus that no longer describes the deployment; a
fixture rewritten under `frontend/tests/fixtures/` touches no service path, and the service job reports
green on a corpus it never read. The contract suite reads the whole tree, on every pull request and on both
publication channels, with no path filter and no draft exemption, which is what makes it the one place a
rule spanning both stacks can be asserted at all. Moving it into either stack would not weaken it — it
would silence it for exactly the changes it exists to catch.

## What it compares

For each fixture, the closure of property names and enumeration spellings reachable from what the route it
answers for records, then the field names and quoted values the fixture states:

- a field spelled the way the contract spells it, in a different case, fails naming the route, the field,
  and both spellings;
- a field the contract records nowhere on that route fails naming the route and the field;
- a quoted value that case-folds to an enumeration the route carries but is not spelled the way the
  contract spells it fails naming the route, the value, and both spellings.

The casing case is the failure this exists for: an enumeration arrives camel-cased over the MCP surface and
Pascal-cased over `/api/client`, and a fixture copying the wrong one passes every gate the client has,
because the hand-written types beside it agree with it.

A route the contract records and no fixture covers is **reported rather than failed** — the run prints
*the corpus states no answer for 19 of the 40 client routes the contract records*, with the list — because
a route nothing has reached yet is work that has not happened rather than a defect. A fixture stating an
answer for a route the contract records no JSON body for fails, and so does a fixture module declaration
that states fields and that no binding names, so a new fixture cannot be added and go silently unchecked.

Which answer each fixture states is a table in the script rather than a marker in the corpus, because
`frontend/tests/AGENTS.md` § *The corpus* refuses a fixture that names a route: the corpus is data, and the
binding is the check's own reading of it. The four values that answer for no route each carry the reason —
the signal payloads travel over a stream the contract records nothing about, the draft listing's envelope
is deliberately not invented, the send refusal is a problem-details body with an extension no OpenAPI
document carries, and the conversation table's `answersRow` is a position in a list rather than a field.

It reads the fixtures as text rather than by loading them, for the same reason this suite fakes `dotnet`
with a symlink to itself: the job installs nothing, and the corpus is TypeScript whose imports carry no
extension. `the_corpus_comparison_names_the_route_the_field_and_both_spellings` holds the comparison itself
to a block carrying one of each kind of disagreement, so a sweep whose extraction stopped matching anything
fails rather than reading as a clean tree.

## What the first run found

`movesPartlyRecorded` had copied the flags result shape. A move records one `change` rather than a list of
them and states the `destinationFolder` it was recorded against, where the fixture wrote `changes` and
`detail` — an answer no deployment produces, sitting in a fixture nothing had read yet. That is fixed here.

## Not in scope

Generating the client's types from the contract, any change to how `http-api-contract.json` is recorded,
and the MCP surface's own contract, each for the reason the issue gives. The check also does not read
whether a fixture *omits* a field the contract requires: a fixture legitimately states the part of an
answer a check reads, and refusing that would be a rule about how fixtures are written rather than about
whether they agree with the deployment.

## Worth knowing

The contract records no failure body for `POST /api/client/drafts/{draftId}/send`, so `refusedSend` — the
shape a client matches a refusal by — is checked against nothing. That is a gap in what the endpoint
declares rather than a defect in the fixture, and it is why that value answers for no route here.

## Verification

`scripts/verify-full.sh` — passed, 294 contracts, 0 failed.

Issue: https://github.com/Krzysztof318/MailFathom/issues/1703


Krzysztof318:
`Frontend / Check, build, and test the client` is red, and it is #1715 rather than anything here: the same three browser specs that fail on `main` itself — `runs nothing the markup carries…`, `draws the region again when it is retried…`, and `draws the mail screens at the phone composition…` — with 45 passing beside them. This change touches no client source and no browser spec; the one fixture it edits, `movesPartlyRecorded`, is imported by nothing (`client.spec.ts` reaches `deployment`, `mail`, and `messages` only). `Required CI` is red because it aggregates that job.

`Repository contracts / Workflow contracts` — the job this change actually adds to — is green, with the two new contracts passing.

`scripts/verify-full.sh` passed locally over this tree: 294 contracts, 0 failed.

Issue #1287, opened 2026-08-25
[P] Run MailFathom in more than one replica, coordinated through PostgreSQL alone


Krzysztof318:
## Context

The Helm chart already exposes `replicaCount` and defaults it to `1`, and nothing refuses a larger number. Part of the service is ready for one: the job queue and the outbox claim a row under `FOR UPDATE SKIP LOCKED` and hold a lease a crash releases, which [ADR 0009](https://github.com/Krzysztof318/MailFathom/blob/main/docs/decisions/0009-durable-job-store-and-execution-identity.md) records; the job schedule lets two replicas reaching one occasion resolve through the queue's own uniqueness; the mailbox refresh-token store lets PostgreSQL settle a concurrent rotation; and the MCP transport is stateless by construction.

The rest is not. A second replica starts a second supervisor for every configured account and a second push-notification watch holding its own IMAP connections, which is what a provider limit or `mail_max_userip_connections` refuses. It starts a second embedding backfill advancing the one deployment-wide resume cursor, a second extraction backfill, and a second stored-content move carrying passes over the same run row. Bounds worded as the deployment's are counted per process and therefore multiply by the replica count: the stored-content ceiling, the embedding request pacer, the answering spend window, and the AI provider concurrency budget. And the administrative surface answers out of a per-process ledger, so what an operator reads describes whichever replica the load balancer picked rather than the deployment.

MailFathom refuses a message broker, a generic workflow engine, and a separate scheduler process until PostgreSQL-backed work demonstrates a concrete limitation. ADR 0009 records that refusal and the shape that answered it. This feature extends the same shape to the work the job store never covered, and it does not reopen the refusal for any of that work.

**One exception has been decided, and it coordinates no work.** #1838 settled on 2026-09-11 how a client's live updates reach it whichever replica serves the connection. They go through an optional RESP pub/sub backplane: any RESP-compatible endpoint the operator already runs, or the server the deployment assets start where they run their own — Garnet today, Valkey once #1917 lands. The reason is that no first-party SignalR backplane runs over PostgreSQL. The backplane carries client signals and nothing else. A deployment running one replica runs none. Every kind of work that must not run twice is still coordinated through PostgreSQL alone. That answer added four children to this parent: #1878, #1879, #1880, and #1881.

## User stories

**As an operator** I want to run three replicas behind one service, **so that** a rolling upgrade and a lost node do not stop mail from being synchronized and questions from being answered.

[...]

PR #1841, opened 2026-09-10
[#1289] Hold work that must not run twice as a lease row in PostgreSQL, with no consumer yet

### LINKED ISSUE #1289: Give the deployment one way to hold work exclusively, as a lease in PostgreSQL with no consumer yet
## Context

The decision this follows records what a lease is; nothing implements one yet. The job store leases a job, but its lease is a stamp on the job row and is reachable only through the job's own claim statement, so a supervisor, a sweep, and a content move have nothing to take. Each of the three lands in a different boundary and none of them should carry a coordination mechanism of its own.

What is missing is one application-owned port for taking, renewing, and releasing exclusive hold of a named unit of work, its PostgreSQL adapter, and — where the decision chose a row over an advisory lock — the migration that adds the table. It ships with no consumer, exactly as the job store did.

## User stories

**As an agent implementing a leased worker** I want one port with one meaning, **so that** synchronization, the sweeps, and the move do not each grow their own idea of what holding work means.

**As an operator** I want a replica that died to release what it held without me acting, **so that** a lost node delays work by a lease period rather than until I notice.

## Acceptance

- An application-owned port takes, renews, and releases hold of a unit named by a bounded, operator-readable key composed of MailFathom's own names — never a subject, an address, or anything from a message.
- The adapter lives in `Infrastructure/Persistence/`, holds PostgreSQL details inside itself, and exposes no provider type across the boundary.
- Where the decision chose a lease row, the change adds one additive migration and no baseline is regenerated; the entity, its configuration, and the snapshot move together.
- Taking is atomic between replicas: two replicas asking for one unit at the same instant leave one holder, settled by PostgreSQL rather than by a check between two statements.
- An expired hold is takeable, and release is conditional on the holder still matching, so a late release cannot free a hold another replica already took.
- Telemetry publishes who holds what and how often a hold changed hands, without recording mail or anything derived from it.
- Unit tests cover taking, contending, renewing, expiring, releasing, and releasing late; the integration suite proves the atomicity against a real PostgreSQL.

## Out of scope

- Any consumer. Synchronization, the sweeps, and the move each take the port in their own child.
- Changing how the job store leases a job. It already has an answer and keeps it.

## Dependencies

- Waits on #1288.

Krzysztof318:
Closes #1289

ADR 0031 chose a lease row over an advisory lock and left the mechanism to be built. This builds it and stops there: one application-owned port for taking, renewing, and releasing exclusive hold of a named unit of work, its PostgreSQL adapter, the additive migration that adds the table, and the telemetry an operator reads it by. Nothing takes it — synchronization, the sweeps, and the move each do that in their own child of #1287 — and nothing changes about how the job store leases a job.

## The port

`Application/Coordination/` is new and holds four types and no dependency on anything below it.

- `WorkScope` is the unit of exclusion, composed of MailFathom's own configured names (`mail-account:personal`, and whatever a later consumer composes), bounded at 256 characters and refusing a blank or control-carrying value.
- `WorkLeaseHolder` identifies one *hold* rather than one process, which is what ADR 0031 asks for: `NewHold()` mints a fresh identity per take, so a replica that lost a lease and took it again is not the holder it was.
- `WorkLease` is what a caller acts on — the scope, the hold, and when it runs out.
- `IWorkLeaseStore` is the port: `ClaimAsync`, `RenewAsync`, `ReleaseAsync`. It takes no persistence session, because a hold outlives the transaction that took it and there must be nothing for a caller to enlist it in.

A claim refuses a live lease whoever asks, including the hold already holding it — a holder renews rather than re-claims. That makes "a granted claim" exactly "a change of holder", which is what lets the claims counter answer how often a hold changed hands without a second instrument to correct it.

## The mechanism is the database's

Three hand-composed statements in `Infrastructure/Persistence/Coordination/`, each one statement because the guarantee *is* the atomicity:

- the claim inserts on the primary key and, where a row is already there, takes it over through `ON CONFLICT … DO UPDATE … WHERE work_leases."ExpiresAt" <= now()`, so two replicas asking at the same instant leave exactly one holder and the loser is refused by PostgreSQL rather than by a read the application made first;
- the renewal and the release are each one conditional update predicated on the holder, so a hold that was already taken over renews nothing and frees nothing — which is what stops a late shutdown handing away a scope another replica is working under.

**Every instant is PostgreSQL's own.** Nothing here reads a clock: the row is stamped with `now()`, judged against `now()`, the duration crosses as an `interval` parameter, and the expiry a caller is handed comes back out of the row through `RETURNING`. A comparison made against each replica's own clock would be decided by the drift between them — a replica running minutes fast would find a live lease expired and take a scope somebody is working under — and ADR 0031 promises the guarantee holds through a clock that drifted.

A release **deletes** the row rather than clearing it, so the table is the set of scopes something is holding: that is what an operator reading it wants, what the gauge publishes, and what lets the next claim take the scope through its insert path rather than waiting out an expiry.

## Migration

One additive migration, `AddWorkLeases`, adding `work_leases` (`Scope` primary key, `Holder`, `HeldSince`, `ExpiresAt`). No baseline regenerated, no existing migration touched.

## Telemetry

`WorkLeaseTelemetry` publishes three instruments — `mailfathom.work_leases.claims` and `.renewals`, each tagged `granted` or `refused`, and an observable gauge `mailfathom.work_leases.held` over the scopes this replica is holding. The scope is a dimension because it is composed of MailFathom's own configured names; the holder identity deliberately is not, being both high-cardinality and answerable from the resource attributes of whichever replica reports the gauge. Nothing records mail.

## Tests

- Unit: the three value types and their guards; the three statements asserted as text through `nameof` so a renamed property fails at compile time rather than at run time — the conflict target, the expiry comparison, both holder predicates, that no `DateTimeOffset` crosses the boundary, and that the scope and hold are parameters; and the telemetry's counters, tags, and gauge.
- Integration (`OrchestratedWorkLeaseStoreTests`, 8 tests): a free scope taken; **eight replicas contending through `ConcurrentIdempotency.RunAsync`, asserted with `AssertSingleEffect`** — which is the atomicity claim, and it is settled by real PostgreSQL rather than by an interleaving a harness chose; a live scope refused; an expired scope taken; a renewal moving the expiry; a renewal after a takeover renewing nothing; a release freeing the scope immediately; a late release freeing nothing.

## Documentation

`docs/architecture/stored-email-schema.md` gains the table and the reasoning behind release-deletes and the single clock; `docs/operations/telemetry.md` gains the three instruments and why the holder is not a dimension; ADR 0031's `describes:` marker now names the files that implement it.

No new dependency, no configuration key, and no change to any of the four public surfaces.


backend/src/Application/Coordination/IWorkLeaseStore.cs
```diff
@@ -0,0 +1,98 @@
...
+/// be lost by a rollback somewhere above it.
+/// </para>
+/// <para>
+/// What a holder is promised is <strong>at most one writer, and never at most one runner</strong>, which
+/// <see cref="WorkLease" /> states in full. A holder that fails to renew stops its own work rather than waiting for the
+/// expiry, and a scope's exclusion reaches MailFathom's own state rather than a connection a mail server has already
+/// accepted.
+/// </para>
+/// </remarks>
+public interface IWorkLeaseStore
```
Krzysztof318, line 33: work lease is store would I consider blocking sql transaction instead of optimistic concurrency

Krzysztof318: It reads as optimistic concurrency but it is not one, and I think that is the whole of the answer: there is no read, no version to compare, and nothing for `OptimisticConcurrencyRetryPolicy` to retry. The exclusion here is a lock — PostgreSQL's, taken inside the statement. `INSERT … ON CONFLICT DO UPDATE` locks the conflicting row when it takes the update path ("all rows will be locked when the `ON CONFLICT DO UPDATE` action is taken"), evaluates the `WHERE` last, against the version the winner left, and guarantees one of insert-or-update "even under high concurrency". So the loser of a race does block; it blocks for the length of one statement rather than for the length of a transaction this code opened.

That is the difference a blocking transaction would make, and it is a change of contract rather than of mechanism. `BEGIN; SELECT … FOR UPDATE; …; COMMIT` turns "tell me whether this scope is free" into "wait until it is", and the waiting costs a connection and an open transaction for as long as the incumbent holds the scope — minutes, by configuration — so the pool would be sized by the number of contended scopes rather than by the work being done. Nothing in #1287's consumers wants to queue behind a scope: a replica that finds an account already being synchronized should move to the next account, which is what an empty claim answer says.

The deeper reason is lifetime, and it is the one ADR 0031 already settled. A lock lives as long as its session, so holding work under one means keeping the transaction open for the whole hold — which is the advisory-lock option that ADR rejected, and it fails in both directions: a partitioned replica loses its lock the moment the connection drops while it is still running the work, and a stalled one keeps it until TCP gives up. A row with an expiry is what makes the hold outlive the session that took it, and what makes "an expired hold is takeable" expressible at all; a `FOR UPDATE` has no expiry to compare against, so the same `ExpiresAt` predicate would have to sit inside the transaction anyway and the lock would be buying only the wait.

Where a blocking transaction is the right shape is where the critical section *is* one transaction — a short decision that must see a consistent row and commit before anyone else reads it. `backend/src/Infrastructure/AGENTS.md` names that case, and it is what a job's claim in ADR 0009 does. This is the other case: the hold is meant to span an account run and outlive every transaction inside it, which is why the port takes no `IPersistenceSession` — there must be nothing for a caller to enlist it in, so that no rollback of somebody else's work can hand the scope away.

Issue #1295, opened 2026-08-25
Ship the deployment contract for more than one replica: the chart, the rollout, and the page that says what scales


Krzysztof318:
## Context

`values.yaml` carries `replicaCount: 1` under a bare `### Workload` heading, with nothing said about what a larger number does. Once the children of #1287 have landed, a larger number is safe — and an operator has no way to learn that from the chart, from `deploy/helm/mailfathom/README.md`, or from `docs/operations/deployment-kubernetes.md`, none of which mentions replicas except to say the PostgreSQL StatefulSet is one on a ReadWriteOnce claim.

Two further things the chart never had to say become true with a second replica. The rollout strategy decides whether an upgrade briefly runs two versions, which is what makes the job queue's handler filter and the leases matter. And a PodDisruptionBudget is what stops a drain taking every replica at once, which for a deployment carrying leases means waiting out an expiry for every account rather than handing them over.

#1838 settled what several replicas need in front of a client. The signal connection is a WebSocket that never negotiates, so the load balancer needs no affinity. It must pass the upgrade and must not close a standing connection sooner than the hub's keep-alive keeps it busy. Signals cross replicas through the backplane that #1880 and #1881 add. The chart already refuses more than one replica serving the client surface without a backplane, so this page is where an operator reads why.

## User stories

**As an operator** I want the chart to tell me what raising `replicaCount` does, **so that** I do not find out from a mail server refusing my connections.

**As an operator** I want a voluntary drain to hand work over rather than park it, **so that** a node upgrade does not stall synchronization for a lease period.

## Acceptance

- `values.yaml` documents `replicaCount` where it is set, saying what more than one replica gives, what it does not, and which bounds are per process; the chart's schema still accepts what it accepted.
- The Deployment carries a rollout strategy the chart states rather than inherits, and a PodDisruptionBudget an operator can turn off, both rendered from values.
- `deploy/helm/mailfathom/README.md` says what the chart renders for more than one replica, keeping every link absolute as that file requires.
[...]

PR #1911, opened 2026-09-11
[#1881] Run Garnet as an optional signal backplane beside the Compose and Quadlet deployments

### LINKED ISSUE #1881: Run Garnet as an optional signal backplane beside the Compose and Quadlet deployments
## Context

The owner settled #1838 on 2026-09-11. The signal backplane is optional in every kind of deployment, not only the chart. Compose and Quadlet run one MailFathom instance on one host today, so a backplane there has nobody to fan signals out to until a second instance runs. The owner chose to ship it now anyway, so that every kind of deployment offers the same component, and so that a later cache can reuse it. Both assets already carry optional services, each in its own idiom. Compose puts them under a profile (`personal-data-scanning`, `spam-scanning`, `object-storage`). Quadlet makes a unit optional by leaving it uninstalled (`mailfathom-presidio.container`).

## User stories

**As an operator on Compose or Podman** I want to switch on the same backplane the chart offers, **so that** my deployment is not a reduced version of the Kubernetes one.

**As an operator who already runs Redis** I want to point the instance at it instead, **so that** nothing more is started.

## Acceptance

- `deploy/compose/compose.yaml` gains a Garnet service under a profile of its own:
  - its image is pinned by digest;
  - it has no volume;
  - its password comes from the secrets directory the file already reads.

  MailFathom reads the connection string from configuration the file already documents. Enabling the backplane means switching the profile on and setting that value. Pointing at an external endpoint means setting the same value without the profile. Without the profile, `docker compose up` brings up exactly the services it does today.
- `deploy/quadlet/` gains a `mailfathom-garnet.container` unit:
  - it is optional by being left uninstalled;
  - it joins the backend network;
  - its image is pinned by digest;
  - its password comes from a secret.

  `mailfathom.container` takes the connection string the same way whether it names that unit or an external endpoint.
- `docs/operations/deployment-compose.md` and `docs/operations/deployment-quadlet.md` say what the backplane does on one host today: nothing, until a second instance runs. They also say how to point at an external endpoint instead.
- `QUADLET_UNIT_DIRS=deploy/quadlet /usr/lib/systemd/system-generators/podman-system-generator --user --dryrun` generates the new unit and reports no unrecognized key.
- The Garnet image's row in `THIRD_PARTY_LICENSES.md` names these assets beside the chart.

## Out of scope

- Running more than one MailFathom instance under Compose or Quadlet. #1295 decides whether that is supported.
- Garnet as a cache, and persistence.

## Dependencies

- Waits on #1879, which defines the configuration these assets set.
- Runs beside #1880, #1878, and #1877.
- #1295 waits on it, because it states what Compose and Quadlet support.

Krzysztof318:
Compose and Quadlet gain the same signal backplane the chart deploys, each in its own idiom for an optional service: a `garnet` service behind a `signal-backplane` profile, and a `mailfathom-garnet.container` unit that is optional by being left uninstalled. `docker compose up` without the profile brings up exactly the two services it does today, and an operator who copies the other unit sources installs nothing extra.

**On one host it does nothing, and both pages say so.** These shapes run one MailFathom, so every signal already reaches every client connection the deployment holds and the server relays each one back to the single subscriber that published it. It is here so that every deployment kind offers the same component and a deployment that grows a second instance has the piece provisioned before it needs one. #1295 is where running more than one is decided.

Pointing at a Redis-compatible endpoint an operator already runs is the other supported shape in both: the connection string is written the same way, and no profile is enabled and no unit is copied.

## How the password reaches the server, and why it differs from the chart

The chart expands the password into `--auth Password --password` from a Secret the kubelet reads. Neither runtime here has that expansion — Compose interpolates `${...}` from `.env` while it parses the file, and podman passes a container's arguments through unchanged — so that route would put the password in `.env`, in the unit file, in `docker inspect`, in `podman inspect`, and in the host's process list rather than in the secrets directory and the credential store beside every other credential.

`--auth ACL --acl-file` is the one mode Garnet reads its credentials from a path in, so both assets start it that way over a one-line Redis-style ACL document granting the default user every command. The default user is what a RESP client authenticating with a password alone becomes, which is what keeps the connection string MailFathom presents identical in all three shapes. The acceptance asks for the password to come from the secrets directory in one and from a secret in the other, and this is the only mechanism the server offers that satisfies either.

## Everything else

- The image is pinned by the same digest the chart renders, Garnet 2.1.0, in both assets. Its release tags are published beside moving `1`, `2`, and `latest` ones, so a digest is the only reference that names one artifact.
- `MailFathom` takes the connection string from lines commented out rather than defaulted — writing the section at all is what says a backplane exists, so an empty reference is refused at startup. That is the same reason the object store's trust anchor is commented out in both files.
- Neither carries a volume, a published port, or a health check. The unit therefore declares no `Notify=healthy` and `mailfathom.container` gains no ordering line for it: MailFathom finishes starting whether or not the endpoint answers.
- The image runs as its own unprivileged account and declares `VOLUME /data`, so both assets give it a memory-backed `/data` beside `/tmp` on a read-only root filesystem, and the unit carries `UserNS=keep-id:uid=1654,gid=1654` so the credential is readable.
- The Garnet image's row in `THIRD_PARTY_LICENSES.md` now names all three assets and records how the password reaches the server in each.

## Verification

- `scripts/test-agent-workflow.sh` — 318 passed, 0 failed.
- `scripts/verify-fast.sh` — reports that the change reaches neither stack, so nothing was restored, built, or tested.
- `docker compose config` renders the file with and without the profile: the default services are unchanged, and the profile adds `garnet` alone.
- `scripts/update-dependencies.sh --only images` sees the two new pins as carriers of the image the chart already names.
- **The Quadlet generator was not run**: podman is not installed on this machine, so `QUADLET_UNIT_DIRS=deploy/quadlet /usr/lib/systemd/system-generators/podman-system-generator --user --dryrun` could not be executed. The unit uses no key that is not already used by a unit in this directory, which is what the check would be asserting, but that is an argument rather than a run.
- The Garnet container itself was not started: this host's daemon is refused by the registry (`error from registry: denied`), so the ACL file, the read-only root filesystem, and the two tmpfs mounts were reasoned from the image's own `Dockerfile` and from what the chart's pod already gives it rather than observed.

Closes #1881


deploy/quadlet/mailfathom-garnet.container
```diff
@@ -0,0 +1,109 @@
...
+# **On one host it does nothing, and that is the expected state rather than a misconfiguration.** This directory runs one
+# MailFathom, so every signal already reaches every client connection the deployment holds and the server relays each one
+# back to the single subscriber that published it. It is here so that this shape offers the same component the Helm chart
+# does, and so that a deployment which grows a second instance has the piece provisioned before it needs one. Running
+# more than one MailFathom under Quadlet is not supported today; issue #1295 decides whether it will be.
+#
+# Rootless, like the units beside it, and installed under ~/.config/containers/systemd/.
+
+[Unit]
+Description=Signal backplane for MailFathom
```
Krzysztof318: Name this just Cache backplane to not link name to signalr only

deploy/compose/.env.example
```diff
@@ -333,6 +333,46 @@
...
+# whether it will be.
+#
+# Enabling it is a profile and a credential rather than a switch, because what says a backplane exists is the
+# configuration section itself. In this order:
+#
+#   1. Choose one password and write it into both files. The first is the server's authority to accept a connection and
+#      is a Redis-style ACL line; the second is the whole connection string MailFathom presents, which the server's own
+#      container name and port are already in:
+#
+#        printf 'user default on >%s +@all\n' "$password" > secrets/garnet-acl
```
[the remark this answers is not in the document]
Krzysztof318: Rejecting this one, with the upstream source as the evidence: `~*` is a documented no-op in Garnet and `&*` is not in its ACL grammar at all, so the proposed correction would stop the server from starting.

Garnet implements a *subset* of the Redis ACL syntax, and `libs/server/ACL/ACLParser.cs` states the whole of it in its own doc comment:

```
ACL_RULE := user <username> (<ACL_OPERATION>)+
ACL_OPERATION := on | off | +@<category> | -@<category>
```

plus the password operations and individual `+command` / `-command` grants. Reading `ApplyACLOpToUser` in that file:

- `~*` and `ALLKEYS` are accepted and deliberately do nothing. The branch carries the comment "No-op, because only wildcard key patterns are currently supported" — key access is all-or-nothing in Garnet and is always granted, which is also why `NetworkGET_SG`'s scatter-gather fast path is safe.
- `&*` matches no branch and falls through to the final `else`, which throws `ACLUnknownOperationException`. An ACL file carrying it fails to parse, and `--acl-file` parses at startup, so the container would crash-loop rather than serve.
- There is no channel permission to grant. `libs/server/ACL/` holds `CommandPermissionSet` and nothing else of the sort, and `User` exposes `CanAccessCommand` and `CanAccessCustomCommand` and no key or channel check. `PUBLISH` and `SUBSCRIBE` are authorized by the command grant alone, which is what `+@all` gives — `@pubsub` is one of the categories `ACLParser` registers, and `@all` covers it.

So `user default on >PASSWORD +@all` grants exactly what the backplane needs, and it is the widest grant the syntax can express.

On the citation: ADR 0032 says the backplane speaks RESP pub/sub, which Garnet answers; it does not claim Garnet implements Redis's key and channel ACL model, and the source above is the authority on that.

Leaving the three commands as they are. Sources: <https://github.com/microsoft/garnet/blob/main/libs/server/ACL/ACLParser.cs>, <https://github.com/microsoft/garnet/blob/main/libs/server/ACL/User.cs>, <https://github.com/microsoft/garnet/blob/main/website/docs/getting-started/security.md>

deploy/quadlet/mailfathom-garnet.container
Krzysztof318: Done in f58061880 — the unit's `Description=` now reads `Cache backplane for MailFathom`, so the container's own name says what it is rather than what the first caller uses it for.

The rest of the vocabulary stays as it is on purpose: the configuration section is `SignalBackplane`, the Compose profile is `signal-backplane`, and both documentation sections are "The signal backplane", because those name the feature the section switches on rather than the server. The unit file, the container, and the Compose service are all `garnet` / `mailfathom-garnet`, which stays neutral if a cache later shares the same endpoint.

PR #1993, opened 2026-09-13
[#1989] Judge a job's lease and due time by PostgreSQL's clock rather than the claiming replica's

### LINKED ISSUE #1989: Judge a job's lease and due time by PostgreSQL's clock rather than the claiming replica's
## Context

A job's lease is stamped and judged against the clock of the replica that asks, while a work lease is stamped and judged against PostgreSQL's `now()`.

- `JobStore.ClaimAsync` reads `TimeProvider.GetUtcNow()` and `JobClaimStatement` compares `"LeaseExpiresAt" <= {claimedAt}` and `"AvailableAt" <= {claimedAt}` against that value, then stamps `"LeaseExpiresAt" = {claimedAt + LeaseDuration}`.
- `JobStore.RenewLeaseAsync` stamps `"LeaseExpiresAt"` from the same process clock and returns an expiry computed beside the row rather than read out of it.
- `WorkLeaseStatements` does the opposite on purpose: every instant is `now()`, the duration crosses the boundary as an interval, and the expiry is read back from the row, so that the exclusion does not depend on drift between replicas ([ADR 0031](https://github.com/Krzysztof318/MailFathom/blob/main/docs/decisions/0031-dividing-singleton-work-between-replicas-with-a-leased-scope.md)).

So two replicas whose clocks disagree judge one job's lease differently. A replica running ahead by more than `Jobs:LeaseDuration − Jobs:ExecutionTimeout` (three minutes with the defaults) reads a live claim as expired and claims the job while its holder is still executing it, which is the duplicated execution the lease exists to prevent. A replica running behind claims an expired job late, and a pending job's `AvailableAt` — its retry backoff — is honoured early or late by the same amount. With NTP the drift is milliseconds and the margin holds; without it, nothing bounds it.

Found while answering whether the database-held locks and leases stay correct against a replicated PostgreSQL cluster. They do in the supported topology (one primary, every session on it); this is the one coordination path that rests on a process clock instead of the database's.

## User stories

- As an operator running several replicas, I want a job claimed by one replica never to be claimed by another while its lease is live, whatever the replicas' clocks say, so that a job runs once per attempt.
- As an operator, I want a retry's backoff to elapse by the database's clock, so that a replica with a skewed clock neither starts it early nor starves it.
- As a maintainer, I want the job queue and the work lease to judge an expiry the same way, so that one coordination rule holds across the service instead of two.

## Acceptance

- The job claim judges `AvailableAt` and `LeaseExpiresAt` against PostgreSQL's `now()`, and stamps `LeaseExpiresAt` as `now()` plus the lease duration passed as an interval.
- Lease renewal stamps `LeaseExpiresAt` from `now()` and returns the expiry read back from the row (`RETURNING`), not one computed in the process.
- Every other comparison in `JobStore` and `DeadLetteredJobStore` that decides whether a job is due or its lease has run out uses the same clock; a timestamp that only records history (`StateChangedAt`, `EnqueuedAt`) may stay on `TimeProvider`, and the choice is stated where it is made.
- Statement-text unit tests assert `now()` in the claim's due predicates and in the renewal, in the way `WorkLeaseStatements` tests already do.
- `Jobs:ExecutionTimeout` stays a process-local duration; its gap to `Jobs:LeaseDuration` covers the attempt's own stopping time and the round trip, not clock skew between replicas.
- No schema change.

Krzysztof318:
Closes #1989

## What changed

Every instant that decides whether a job is due or its lease has run out is now PostgreSQL's `now()`, with durations passed as intervals — the rule `WorkLeaseStatements` already follows. Instants that only record history (`EnqueuedAt`, `StateChangedAt`) stay on `TimeProvider`, and `JobStore`'s remarks state that split where it is made.

- **Claim** (`JobClaimStatement`): both due predicates compare `AvailableAt` and `LeaseExpiresAt` with `now()`, and the claim stamps `LeaseExpiresAt = now() + interval`.
- **Renewal** (new `JobLeaseRenewalStatement`): stamps `LeaseExpiresAt = now() + interval` and returns the expiry read back from the row with `RETURNING`, rather than one computed in the process.
- **Retry backoff**: `IJobStore.ScheduleRetryAsync` now takes the retry **delay** rather than an absolute instant. It writes `AvailableAt = now() + delay`, moves `TurnAt` forward to at least that instant, and returns the stored `AvailableAt`, or `null` when the attempt no longer held the job. `JobExecutor` reports that stored instant as `NextAttemptAt`.
- **Release** and **returning a dead letter** (`DeadLetteredJobDecisionStatements.ComposeRetry`): `AvailableAt` (and the returned dead letter's `TurnAt` floor) is `now()`.
- **Enqueue** (`JobEnqueueStatement`): a job that names no available instant is available at `COALESCE(<named instant>, now())`. An instant the enqueuer names is kept as given, because it is the enqueuer's own schedule.

`Jobs:ExecutionTimeout` stays a process-local duration. `docs/operations/configuration-runtime.md` now says its gap to `Jobs:LeaseDuration` covers stopping the attempt and the round trip, not clock drift. `docs/architecture/stored-email-schema.md` states which clock each job instant is judged by.

## Public surfaces

No break. No configuration key, database schema, MCP tool, or HTTP contract changes, and no migration is needed. `IJobStore` is an internal application port; its single implementation and its test doubles change with it. During a rolling upgrade, an older replica still stamps due and lease instants from its own clock, which is exactly the behaviour today, so the mixed window is no worse than the current state.

## Tests

- Statement-text unit tests assert `now()` in the claim's due predicates and lease stamp, in the renewal's stamp and `RETURNING`, in a returned dead letter's available instant and turn, and in an enqueue that names no instant. They also assert the lease duration reaches the statement as a parameter and no process instant decides anything.
- `JobExecutorTests` pass the delay and assert the reported `NextAttemptAt` is the instant the store returned.
- `OrchestratedJobStoreTests.ScheduleRetryAsync_ByTheHolder_HoldsTheJobBackUntilTheInstantItNames` now passes a one-hour delay. That suite runs only in the pipeline, so it was not run locally.

## Not in this change

The outbox's own lease (`OutgoingEmailClaimStatement` and the conditional writes in `OutboxOperationStore` / `OutgoingEmailStore`) still stamps and judges `AvailableAt` and `LeaseExpiresAt` from the process clock, so it has the same skew exposure. It is outside this issue's acceptance, which names `JobStore` and `DeadLetteredJobStore`.


Krzysztof318:
Answering the review of `156fd97`, whose only finding had no line to sit on:

- **P2 — Update the timeProvider parameter doc for the retry instant** (`backend/src/Application/Jobs/Execution/JobExecutor.cs`): fixed in `251997d`. The `timeProvider` parameter doc now says the clock times the attempt, the timeout, and the renewal interval, and that the next attempt's instant is stamped by the store. The executor no longer reads its clock for that instant; it passes the backoff delay, and the store writes `now() + delay`.

The branch was also rebased onto the current `main`.

Issue #2068, opened 2026-09-19
Grow the structured-agent evaluations to fifteen cases each, and the relevance filter's labelled set to thirty


Krzysztof318:
## Context

The structured agents under #2043 are measured on three to seven cases each, and `email-enrichment` on exactly one message, `CorpusMessage.At(3)`. The relevance filter's labelled set holds seven candidates. At that size a model's result is decided by one case, and a change to an instruction cannot be told apart from noise. The structured answers are what make these agents cheap to measure — most of what matters is a plain assertion that costs no judge call — so a larger set costs little beyond the model calls themselves.

## User stories

**As the owner** I want each structured agent measured on enough cases that one case does not decide a model's result, **so that** which model these frequent calls run on is a number rather than an opinion.

**As the owner** I want the enrichment agent measured on more than one message, **so that** its result says something about enrichment rather than about that message.

## Acceptance

- `email-enrichment`, `thread-state`, `contact-relationship`, and `search-phrase` each have at least fifteen cases.
- `email-enrichment` covers a message with no reading worth writing, one with several amounts and dates, one whose request sits in quoted history rather than in the new text, and a newsletter or notification.
- `thread-state` covers a commitment later withdrawn, a question answered several messages later, and an agreement one side qualifies.
- `contact-relationship` covers a correspondence that changed tone over time and one where the person only ever appears in copy.
- `search-phrase` covers a sentence stating a relative period, one naming two senders, and one mixing a filter with free words.
- The relevance filter's labelled set holds at least thirty candidates, with near misses beside the clearly relevant and clearly irrelevant ones, and its threshold report is read over the larger set.
- Every case asserts plainly what a structure settles; the judge is used only where a case reads two ways, as #2057 established.
- `docs/operations/local-development.md` § *Agent evaluations* reflects the sets.

## Out of scope

- `body-cleanup`, whose evaluation was exploratory.
- The prose-answering agents, which #2067 holds.
- Any change to an agent's instruction.

## Dependencies

Nothing open. #2042 and #2057 delivered the scenarios these cases extend; both have closed.

PR #2082, opened 2026-09-19
[#2077] Keep a search phrase reading whose address filter names two addresses

### LINKED ISSUE #2077: Keep a search phrase reading whose address filter names two addresses
## Context

`MailSearchPhraseDocument.Filters.SenderAddress` is a single string. When a sentence names two senders, all three models the `AI evaluations` run measured for #2068 answered with an array instead, as in the `TwoSenders` case:

```json
{"filters":{"senderAddress":["billing@northwind.example","accounts@fabrikam.example"]},"criteria":["renewal"]}
```

`MailSearchPhraseDocumentReading.ReadDocument` deserializes the whole answer through the source-generated context. The array fails to bind to the string, `JsonSerializer` throws, and the catch returns `null`. So the reading becomes `MailSearchPhraseReading.Nothing`: the criterion `renewal` is lost with the filter, and the search falls back to the sentence's own words. One field the reading cannot use loses the whole answer, including the parts it could have kept.

The instruction says `senderAddress` takes a whole mail address, but it does not say what to write when a sentence names more than one. The `TwoSenders` case expects no sender filter, because the one filter there is would hide the other sender's mail.

## User stories

**As a person searching my mail** I want a sentence naming two senders to keep the words I typed, **so that** the search still ranks by what I described rather than falling back to the raw sentence.

**As the owner** I want one unreadable field to cost that field and nothing else, **so that** a model that answers mostly right is used for what it got right.

## Acceptance

- An answer whose `senderAddress` or `recipientAddress` is not a single string keeps its other filters, its criteria, and its unaccounted text. The address filter it could not read is dropped, and a unit test covers an array in each field.
- The instruction states what to write when a sentence names more than one sender or recipient, consistent with the `TwoSenders` case expecting no sender filter. That change goes through the agent's own instruction and its evaluation, not through this reading alone.
- A dispatched `AI evaluations` run shows `MailSearchPhrase.TwoSenders` passing under the declared models, or the pull request names the model that still falls short and why.

## Dependencies

None.

Krzysztof318:
## What changed

A sentence naming two senders was answered by every model measured for #2068 with both addresses in one field:

```json
{"filters":{"senderAddress":["billing@northwind.example","accounts@fabrikam.example"]},"criteria":["renewal"]}
```

`senderAddress` is one string, so the array failed to bind, `JsonSerializer` threw, and the reading became
`MailSearchPhraseReading.Nothing`. One field the reading could not use cost the whole answer — the criterion `renewal`,
every other filter, and the part the model said it made nothing of — and the search fell back to the sentence's own
words.

Now a text field written in a shape this build cannot use costs that field and nothing else, and the instruction says
what to write when a sentence names more than one sender or recipient: no address filter at all, because either address
written alone hides the other sender's mail.

## How

- `WrittenTextJsonConverter` reads a field as text where a model wrote text and skips past the whole value otherwise, so
  an array, an object, or a number is read as no value rather than as no document. `reader.Skip()` rather than one
  `Read()` is what leaves the reader where the next field begins instead of inside the one that was dropped.
- Every text field of the answer carries it — `senderAddress`, `recipientAddress`, `receivedFrom`, `receivedTo`, and
  `unaccounted` — rather than the two the defect was measured on. They are one field type with one failure, and each
  costs an attribute; a field dropped for its shape then joins the day no calendar holds and the filter this build has no
  chip for, which `MailSearchPhraseDocumentReading` already read that way.
- The instruction gains the rule the `TwoSenders` case was already written against: neither address field takes two
  addresses, and a sentence naming more than one sender or recipient omits the field and leaves what it named in
  `criteria`. That is the agent's own contract rather than the reading's, which is why it moves here as well as there.
- Nothing about validation moved. An address is still kept only where it could be one, a day only in the one shape the
  instruction asks for, and an answer nothing survives of is still the plain word search a deployment with no model runs.

## Tests

- `MailSearchPhraseDocumentReadingTests` covers an array in each address field, an object and a number in one, and the
  two other text fields written as arrays — each asserting that the rest of the answer, filters and criteria and
  unaccounted part alike, is read exactly as it was written.
- `MailSearchPhraseInstructionsTests` holds the instruction to stating that rule, the same way it holds it to every bound
  the reading enforces.
- `MailSearchPhraseScenarioTests` gains the answer from the issue as a passing row for `TwoSenders`: the array answer now
  reads, carries no sender filter, and keeps its criterion, which is what the case asks for. It calls no provider.

## The evaluation run

The instruction text changed, so every cached answer for this agent is a miss and an `AI evaluations` run pays for the
phrase cases again; the other agents' cases stay cached. A run dispatched on this branch is what says whether
`MailSearchPhrase.TwoSenders` now passes under the declared models, and its verdict is reported on this pull request.

## Not in this change

`unread`, `flagged`, `hasAttachments`, and `criteria` are not text and keep their declared shapes, so an answer writing
one of them in another shape still loses the whole document. No measured run has produced one, and each would need a
converter of its own rather than an attribute — worth doing when a run shows it, not before.

Closes #2077


Krzysztof318:
## The evaluation run

`AI evaluations` run [35477588021](https://github.com/Krzysztof318/MailFathom/actions/runs/35477588021), on this
change's merge commit `83d918e1f`, under `deepseek/deepseek-v4.1-flash`, `openai/gpt-5.6-luna`, and
`openai/gpt-oss-120b`, one repetition each.

**`MailSearchPhrase.TwoSenders` passes under all three models**, and so does every other
`MailSearchPhraseEvaluations` case in that run — the suite's twenty failures are in `MailAnswering`,
`ContactRelationships`, `ImageDescription`, `DiscoveryComposition`, `DiscoveryEndToEnd`, `EmailEnrichment`,
`ReplyDrafts`, and `ThreadStates`, none of which this change touches.

One thing an operator has to know before dispatching that workflow again: the repository variable
`EVALUATION_MODELS` is currently empty, and a dispatch that names no `models` input therefore fails every scenario in
the suite at once with `The evaluations were turned on through MAILFATHOM_AI_EVALUATIONS and
MAILFATHOM_EVALUATION_MODELS is not set, so this fails rather than skipping` — run
[35477030777](https://github.com/Krzysztof318/MailFathom/actions/runs/35477030777) is that failure and says nothing
about any model. Either set the variable or name the models on the dispatch.

